PR #10, opened 2020-11-25
Create diseaseModel

UlrikeS91:


UlrikeS91:
@lzehl, could you check if this would make sense like that? The general disease model doesn't have an `ontologicalIdentifier`, but separate model lines could have. Of course, the disease itself has an identifier, which is probably smart to reference at least in the description. Question: Would it make sense for this special case to add a property for this exact purpose? For example, `diseaseOntologyIdentifier`?

Issue #15, opened 2020-12-14
rename wild_type into wildType


olinux:
To be consistent, we should follow the lowerCamelCase syntax for everything in JSON(-LD).

https://github.com/HumanBrainProject/openMINDS_controlledTerms/blob/2400e43e873fbfadd9f776fa69bc541a86144c8a/terminologies/phenotype/wild_type.jsonld#L2

lzehl:
@olinux : I agree! 

@UlrikeS91 : do you agree as well? why was it changed to "wild_type"?

UlrikeS91:
Hmm, good question. I think the logic I followed was that the name is "wild type", so the jsonld should indicate the space. Not sure, why this was the most logical conclusion anymore...

Feel free to change it to `wildType`!

olinux:
fixed

PR #18, opened 2020-12-15
Follow naming conventions in id

olinux:
Renaming wild_type to wildType for the id to follow the naming conventions -> fix for #15

### DIFF
--- a/terminologies/phenotype/wild_type.jsonld
+++ b/terminologies/phenotype/wild_type.jsonld
@@ -1,5 +1,5 @@
 {
-  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wild_type",
+  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wildType",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Phenotype",
   "name": "wild type",
   "definition": null,
--- a/terminologies/phenotype/wild_type.jsonld
+++ b/terminologies/phenotype/wild_type.jsonld
@@ -1,7 +1,7 @@
 {
-  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wildType",
+  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wildtype",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Phenotype",
-  "name": "wild type",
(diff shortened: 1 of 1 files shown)


lzehl:
@UlrikeS91 : I agree with Oli's suggestion (cf. also #15) Do you agree? if yes please feel free to merge.

UlrikeS91:
@olinux and @lzehl, shouldn't the file name also be renamed to "wildType" then?

lzehl:
@UlrikeS91 no, the entries for the "name" property stays at it is usually written. the ID and the type have conventions:
- ID: lowerCamelCase
- type: UpperCamelCase
Which is true for spaces, but only question here is what to do with dashes in names? Should they remain in the ID and type?
@olinux what do you think?

olinux:
I think dashes should be generally avoided if anyhow possible. Otherwise, I would suggest to treat them just as other "filling" characters and therefore treat them as if they would be a space... meaning "foo-bar" becomes "fooBar"

lzehl:
@olinux okay. we can follow this convention. For this PR : @UlrikeS91 is right, but I misunderstood her. Could you change the filename of this json-ld also to "wildType.jsonld" ?

terminologies/phenotype/wild_type.jsonld
```diff
@@ -1,5 +1,5 @@
 {
-  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wild_type",
+  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wildType",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Phenotype",
   "name": "wild type",
```
lzehl: I've checked again. It's completely valid to use the notation "wildtype"
@olinux again, not your mistake but could you fix it ? and change the at_id as well?

terminologies/phenotype/wild_type.jsonld
```diff
@@ -1,5 +1,5 @@
 {
-  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wild_type",
+  "@id": "https://openminds.ebrains.eu/controlledTerms/phenotype/wildType",
```
lzehl: see below (it's okay to use "wildtype" notation)

lzehl (changes requested):

PR #19, opened 2020-12-23
Updating software-related terminologies

jagru20:
- Added schemas for a few more terms.
- Added Links to ontology for existing Terms where possible.


terminologies/contributionType/softwareSupport.jsonld
```diff
@@ -1,7 +1,7 @@
 {
-  "@id": "https://openminds.ebrains.eu/controlledTerms/contributionType/support",
+  "@id": "https://openminds.ebrains.eu/controlledTerms/contributionType/softwareSupport",
   "@type": "https://openminds.ebrains.eu/controlledTerms/ContributionType",
-  "name": "support",
+  "name": "softwareSupport",
   "definition": null,
   "description": "Contributing by providing support to users of the product.",
```
lzehl: If you rename it to softwareSupport the description needs to be specific as well. And we need to introduce other specific support contribution as well (e.g., technicalSupport, scientificSupport?, etc).

lzehl (changes requested):
@jagru20 thanks a lot for updating / extending the software parts in the controlledTerms. It's highly appreciated :slightly_smiling_face: 

A general feedback point: please check again all "names" in your changes. The entries here should be with space and everything. The lowerCamelCase convention needs to be only followed for the properties, and the UpperCamelCase convention for the types. But the values can / should be written normally.

Please, see also my comment for the support.json-ld

lzehl (changes requested):
@jagru20 I think there was a misunderstanding between us for the "name" entries. They do not need to use lowerCamelCase or UpperCamelCase, but they should be written in lower capitals except for abbreviations (e.g. HPC; although abbreviations should only be used in "name" if they are more commonly known than the full name) and given names (e.g., Java). Could you update this accordingly?

Otherwise this PR looks good.

jagru20:
> 
> 
> @jagru20 I think there was a misunderstanding between us for the "name" entries. They do not need to use lowerCamelCase or UpperCamelCase, but they should be written in lower capitals except for abbreviations (e.g. HPC; although abbreviations should only be used in "name" if they are more commonly known than the full name) and given names (e.g., Java). Could you update this accordingly?
> 
> Otherwise this PR looks good.

Yes, I got that wrong then. I will update that.

lzehl:
@jagru20 top :) thanks a lot!

lzehl (approved):

PR #20, opened 2021-02-03
Update termSuggestion.schema.tpl.json

lzehl:
Just a small bug-fix (mismatch between required and defined property) for suggested term.


lzehl:
@olinux could you double check this PR and merge it if it's okay?

olinux (approved):

PR #28, opened 2021-02-26
Terminology update

UlrikeS91:
This was somehow oversee, but should actually reflect the terminologies in `controlledTerms/terminologies` (without `controlledTerms/terminologies/terminology` of course)

@lzehl, especially look at metaDataModelTyp.jsonld if the `"name": "metadata model type"` is ok or should be changed. Preferably also cross-check if I actually added all terminologies (except for `terminology`) ;)  

@olinux FYI


terminologies/terminology/metaDataModelType.jsonld
```diff
@@ -0,0 +1,8 @@
+{
+  "@id": "https://openminds.ebrains.eu/controlledTerms/terminology/metaDataModelType",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Terminology",
+  "name": "metadata model type",
```
lzehl: (meta)data model type

lzehl: to explain: it covers both metadata model types (e.g. openMINDS) AND data model types (e.g. BIDS)

terminologies/terminology/softwareFeatureCategory.jsonld
```diff
@@ -0,0 +1,8 @@
+{
```
lzehl: softwareFeatureCategory should be removed (for all: schemas, terminologies, etc) was depricated

lzehl (changes requested):
@UlrikeS91 thank you :slightly_smiling_face: it just needs some small changes.

lzehl:
@UlrikeS91 and I've double checked: you've got all :wink:

terminologies/terminology/metaDataModelType.jsonld
UlrikeS91: That's what I thought but didn't trust myself :D

terminologies/terminology/softwareFeatureCategory.jsonld
UlrikeS91: If it is depricated, why have the terms in the terminology-folder been updated by @jagru20 3 days ago? 

The schema for this has been deleted already and isn't used anywhere anymore. So, @lzehl is of course right. It is depricated. @jagru20, since you've done recent updates to this depricated terminology, do you need these terms to be moved to ("controlledTerms/terminologies/softwareFeature")? If you don't need them at all, can we delete the entire folder ("controlledTerms/terminologies/softwareFeatureCategory") with all its content?

UlrikeS91:
Changed `"name": "metadata model  type"` to `"name": "(meta)data model  type"`, as requested. Deleted depricated "softwareFeatureCategory.jsonld" in "controlledTerms/terminologies/terminology". 

@lzehl, waiting for feedback from @jagru20 about what to do with "controlledTerms/terminologies/softwareFeatureCategory"-folder - delete or move terms to "controlledTerms/terminologies/softwareFeature".

terminologies/terminology/softwareFeatureCategory.jsonld
jagru20: The only reason why I made changes (i.e., changing the case of name property to lower case) was that the folder was still existent in my fork of the project. It is in deed deprecated and can be deleted.

If someone reads this later and wonders; The decision not to use the softwareFeatureCategory was drawn in https://github.com/HumanBrainProject/openMINDS_core/issues/142

UlrikeS91: Okay, then I'll delete the whole thing :)

UlrikeS91:
Ready for second review. Deleted all remains from "softwareFeatureCategory" within this PR now.

lzehl (approved):

lzehl:
@UlrikeS91 thanks a lot :+1:

PR #49, opened 2021-08-16
adding new techniques (part 2)

lzehl:


lzehl:
@UlrikeS91 this is ready now for a review :)

apdavison (commented):
Do you also plan to add "small-scale simulation" and "medium-scale simulation"? How will the line be drawn?

How about just "simulation"? Or "simulation" + "HPC simulation" if you think the distinction of scale is important.

lzehl:
@apdavison We can yes. For now I was just focusing on terms that were already present in some way in the old system. We need to go as deep as it would make sense for queries. I think large/medium/small makes sense. is large-scale not a synonym to HPC? or are there large-scale simulations that can run without an HPC?

instances/technique/nonrigidMotionCorrection.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/nonrigidMotionCorrection",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": null,
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "nonrigid motion correction",
```
UlrikeS91, line 11: This is lacking the "-" between "non" and "rigid" (or a space, but I think that would be weird). The `@id` need to be adjusted accordingly.

instances/technique/nonrigidRegistration.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/nonrigidRegistration",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": null,
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "nonregid registration",
```
UlrikeS91: same comment as above; additionally, "rigid" has a typo ("r**E**gid")

instances/technique/rigidRegistration.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/rigidRegistration",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": null,
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "regid registration",
```
UlrikeS91: "rigid" with typo again ("rEgid")

UlrikeS91 (changes requested):

apdavison:
Hi Lyuba, I think "small", "medium" and "large" are problematic, because there is no agreed upon definition for how to define the scale of a simulation (number of neurons, number of synapses, volume of tissue simulated, number of processor-hours required...?), and for sure the boundaries change with time (what was large-scale 10 years ago is now medium- or small-scale).

I suggested "HPC simulation" as a more concrete, well-defined term that is approximately equivalent to the intent of "large-scale", but really I would be inclined to just have "simulation".

instances/technique/nonrigidMotionCorrection.jsonld
lzehl: https://www.merriam-webster.com/dictionary/nonrigid should be correct

instances/technique/nonrigidRegistration.jsonld
lzehl: cf above

instances/technique/rigidRegistration.jsonld
lzehl: cf above

lzehl:
@apdavison just "simulation" seems too vague for me to be of real use. I've entered "large-scale simulation" because it has a meaning to people from the field (discussed it with some NEST people), but HPCsimulation would make them also happy (I guess). Could you have a look at https://search.kg.ebrains.eu/instances/069e4718-2b4a-4056-9c8f-e646a841b2dd and tell me which controlled techniques would you filter out of the given methods? (I've kept now "communication profiling", "large-scale simulation", "quantitative analysis" and "qualitative analysis")

UlrikeS91:
> @apdavison just "simulation" seems too vague for me to be of real use. I've entered "large-scale simulation" because it has a meaning to people from the field (discussed it with some NEST people), but HPCsimulation would make them also happy (I guess). Could you have a look at https://search.kg.ebrains.eu/instances/069e4718-2b4a-4056-9c8f-e646a841b2dd and tell me which controlled techniques would you filter out of the given methods? (I've kept now "communication profiling", "large-scale simulation", "quantitative analysis" and "qualitative analysis")

The PR is basically ready, but @lzehl and I would like to wait for @apdavison feedback about the question around the "simulation" technique. @apdavison would you be able to have a look and tell us which techniques would be suitable?

apdavison:
Having now reviewed the level of detail in the other "technique" instances, I definitely think that plain "simulation" would be best. If you want more specificity, then I would put both "HPC simulation" and "neuromorphic simulation" for that specific dfataset. "Large-scale" should be avoided, as the term is subjective and doesn't have a stable meaning across the field or across time (i.e. different people have a different idea about what it means, and people's idea of its meaning changes as computing technology improves).

I would be inclined to remove "communication profiling", as this is a technique from computer engineering, not from neuroscience.

lzehl:
@apdavison and @UlrikeS91 I've renamed "large-scale simulation" to "HPC simulation", and added "simulation" and "neuromorphic simulation" as well. I will leave "communication profiling" because it is still a technique. Note that what we register under techniques is not purely neuroscience, but all techniques that we need to mention for the data we register in the KG.

lzehl:
I've also updated the image registration techniques and the transformation techniques according to the feedback I received from Timo.

instances/technique/imageRegistration.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/registration",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "An 'image registration' is a process of bringing a set of images into the same coordinate system.",
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "transformation",
```
UlrikeS91: `@id` vs. `name`: registration or transformation? Definition points to registration, but also says "image registration" which is yet another thing.

UlrikeS91 (changes requested):

instances/technique/imageRegistration.jsonld
lzehl: screw up on my side... should be corrected now

UlrikeS91 (approved):

PR #52, opened 2021-09-03
Reassessment DataTypes

lzehl:
based on discussion with Stefan we want to reduce the complexity here to not over-complicate the assignment to ContentTypes and Files


skoehnen (approved):
Looks good

skoehnen (approved):

PR #54, opened 2021-09-06
Tissue block as TissueSampleType option

UlrikeS91:


lzehl (commented):
no definition yet? how about "A 'tissue block' is a cube-like part of an organ that was excised from a living or deceased human or animal body for further studies/preparations." ?

lzehl:
@UlrikeS91 what do you think?

UlrikeS91:
> no definition yet? how about "A 'tissue block' is a cube-like part of an organ that was excised from a living or deceased human or animal body for further studies/preparations." ?

Sorry. I forgot to add one yesterday. Your definition isn't quite correct. The tissue sample doesn't need to be cube-like - it's the paraffin wax that is. Also, for small animals, samples can be whole organs (e.g. a whole mouse brain). 

My proposal would be: 
"A 'tissue block' is tissue sample from a living or deceased human or animal body embedded in paraffin wax for further studies/preparations."

The part I'm a bit unsure is whether a tissue block ALWAYS is embedded in paraffin wax or if there are other substances that can be used. Maybe rather something more generalized. Like this:

"A 'tissue block' is tissue sample from a living or deceased human or animal body embedded in a rigid medium (often paraffin wax) for further studies/preparations."

@lzehl, what do you think?

lzehl:
@UlrikeS91  I don't think that all tissue blocks are embedded in paraffin... I was working on tissue blocks that were not embedded...

lzehl:
Let's maybe bring this up today with @tgbugs ?

UlrikeS91:
> Let's maybe bring this up today with @tgbugs ?

Yes! Sounds good.

UlrikeS91:
related to #55 
I'll close this PR for now until we came to a conclusion in the linked issue

PR #61, opened 2021-11-01
add synonyms for units

olinux:
Please validate (especially the "embryonic day" :) )

### DIFF
--- a/instances/unitOfMeasurement/kilobyte.jsonld
+++ b/instances/unitOfMeasurement/kilobyte.jsonld
@@ -10,5 +10,5 @@
   "knowledgeSpaceLink": null,
   "name": "kilobyte",
   "preferredOntologyIdentifier": null,
-  "synonym": null
+  "synonym": ["KB"]
 }
--- a/instances/unitOfMeasurement/kilobyte.jsonld
+++ b/instances/unitOfMeasurement/kilobyte.jsonld
@@ -9,6 +9,8 @@
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
   "name": "kilobyte",
+  "placement": null,
   "preferredOntologyIdentifier": null,
+  "preferredSymbol": "KB",
   "synonym": ["KB"]
 }
--- a/instances/unitOfMeasurement/bit.jsonld
+++ b/instances/unitOfMeasurement/bit.jsonld
@@ -12,5 +12,5 @@
   "placement": null,
   "preferredOntologyIdentifier": null,
   "preferredSymbol": "b",
-  "synonym": ["b"],
+  "synonym": ["b"]
 }
--- a/instances/unitOfMeasurement/bit.jsonld
+++ b/instances/unitOfMeasurement/bit.jsonld
@@ -10,5 +10,5 @@
   "knowledgeSpaceLink": null,
   "name": "bit",
   "preferredOntologyIdentifier": null,
-  "synonym": null
+  "synonym": ["b"]
 }
--- a/instances/unitOfMeasurement/bit.jsonld
+++ b/instances/unitOfMeasurement/bit.jsonld
@@ -9,6 +9,8 @@
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
   "name": "bit",
+  "placement": null,
   "preferredOntologyIdentifier": null,
-  "synonym": ["b"]
+  "preferredSymbol": "b",
+  "synonym": ["b"],
 }
--- a/instances/unitOfMeasurement/byte.jsonld
+++ b/instances/unitOfMeasurement/byte.jsonld
@@ -9,6 +9,8 @@
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
   "name": "byte",
+  "placement": null,
   "preferredOntologyIdentifier": null,
+  "preferredSymbol": "B",
   "synonym": ["B"]
 }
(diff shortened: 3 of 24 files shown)


UlrikeS91:
The abbreviation for "embryonic day" is correct. @olinux, in this case the unit comes before the value, e.g. E11.5 as done for this dataset: https://search.kg.ebrains.eu/instances/Dataset/8ab25629-bdac-47d0-bc86-6f3aa3885f29. 

Generally, it looks good but we should maybe consider if we are happy with the (other) time unit abbreviations. 

unit | abbreviation 
---- | --------------
millisecond | ms 
second | s
minute | min
hour | h
day | d
week | wk
month | mth
year | yr

Everything up to "day" looks correct to me and I think are pretty standard, but what about "week", "month" and "year"? I guess you choose abbreviations that are user-friendly and frequently used, but e.g. I think a more formal abbreviation for "year" would be "a" (as in _anno_). 
Maybe we should involve native speakers, @apdavison or @tgbugs maybe?

olinux:
Thanks for pointing this out. this is tricky with the prefixed "E" - since we need to pass this information somehow to make it possible to be interpreted by the clients... This again speaks for a dedicated property which allows to add information such as a boolean "prefix" or "postfix". Same thing goes for information on how to build the plural which is specific by instance... :sweat_smile:

olinux:
For the formality of the abbreviations: We should think about what the use-cases are. For us it is to show a short form in a UI (here, the "yr" would be preferable to "a") but I do agree that this highly depends on the clients requirements...

tgbugs:
The abbreviations should probably be aligned with other conventions. See for example what I tried to do with https://github.com/tgbugs/protc/blob/master/protc-lib/protc/units/si-units-extras.rkt. One of the issues with time units is that some of the abbreviations have nasty collisions with other official SI units. If you follow the strftime formatting conventions there are also some collisions but at least there is some established convention.

If you want a peek at some of the complexity of trying to disambiguate and parse this stuff see https://github.com/tgbugs/parsercomb/blob/d90511a3a9bf060bc58757215812e33b90cb7e53/pysercomb/parsers/units.py#L309-L362

Frankly, unless there is a very good reason to display abbreviations, I would just leave them out entirely.

instances/unitOfMeasurement/kilobyte.jsonld
```diff
@@ -10,5 +10,5 @@
   "knowledgeSpaceLink": null,
   "name": "kilobyte",
   "preferredOntologyIdentifier": null,
-  "synonym": null
+  "synonym": ["KB"]
```
apdavison, line 13: should be kB ([Wikipedia](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&cad=rja&uact=8&ved=2ahUKEwiCqO66p_nzAhWwxIUKHbSRAoIQmhN6BAhBEAI&url=https%3A%2F%2Fen.wikipedia.org%2Fwiki%2FKilobyte&usg=AOvVaw11-EA9rHiNzVp3r_LYsIhC))

apdavison, line 15: I think we should specify somewhere whether 1 kilobyte is 1000 bytes (1 kB) or 1024 bytes (1 KB), and similarly for MB, etc. I wonder if we should explicitly use kibibyte (KiB) for 1024 bytes?

UlrikeS91: @olinux is going to like you very much for this comment. Seems like everything is pointing to an extension for the unitOfMeasurement schema with properties describing:
- how units can be converted (e.g. msec to sec to min)
- what type of unit they are (e.g. time, size, volume, etc.)
- whether it is a postfix or prefix (e.g. embryonic day as E11.5 vs. day as 11.5 d) 
- which symbol should be used (instead of the workaround to add it under synonyms & with plurals defined if needed?) 

@lzehl, any thoughts on this?

apdavison: the 1000 vs 1024 bytes question would probably be disambiguated by the `knowledgeSpaceLink` or `preferredOntologyIdentifier`, e.g.:

kilobyte (1000 bytes): http://purl.obolibrary.org/obo/UO_0000234
kibibyte (1024 bytes): http://purl.obolibrary.org/obo/UO_0000245

lzehl: As for kilobytes vs kibibytes: both should be units we should have registered (please add kibibytes if they are missing from the controlledTerms list for measurement units).  

For general conversion between units: if we stick to the official names / abbreviations there are tools out there that deal only with unit conversions (time is for me a special issue, though; @apdavison did you ever deal with time conversion? (year -> months -> days; or the other way around). I would rather support the convention used by one or a selection of tools that do correct conversions than inventing a schema that provides information on how to convert in a machine-readable way. There speaks nothing against extending the definitions with such information, though.

There speaks also nothing against extending the controlledTerm schema for UnitOfMeasurement with additional properties. For example, "preferredSymbol" (for openMINDS preference; which is different to synonyms which could list all other stuff), "placement" (with selection prefix/postfix). Plurals should be defined elsewhere (e.g. in ontologies) or we just ignore them (because for display I would anyway only use the preferred symbol which does not have a plural).

lzehl:
@apdavison & @olinux where are we on this PR?

UlrikeS91:
Since this has not been pushed any further and I don't think that there is a strong need to add this, I will close this PR.

Issue #64, opened 2021-11-03
Stimulation approach / Stimulus type: add visual


eapapp:
Hello Lyuba and Ulrike,

I'm in the process of migrating a [dataset](https://editor.kg.ebrains.eu/instances/2077efac-09f6-47c1-aabe-632e08ed148b) in which visual stimuli have been applied, but the KGE only lists auditory stimulation and direct current as possible options for Stimulation approach / Stimulus type. Could you please add visual stimulation? For stimulus type, the authors describe "visual target stimulus (CS+) and distracter stimulus (CS−)" in their [paper](https://www.nature.com/articles/ncomms15602), if that makes sense to add.

All the best,
Eszter

UlrikeS91:
Thanks for the request @eapapp! As you see from the draft PR, the stimulationApproach will be added. 

The stimulusType is a bit more difficult. I had a look at the paper you linked. Here a short summary to save others time:

Rats were trained to perform a two-choice visual discrimination task on a figure-8 maze (an elevated platform that is shaped like an 8). At one end of that maze are two screens where two visual stimuli were presented simultaneously (every time both). The stimuli were black and white images (a diamond and a plane), equiluminant and with same proportions of black/white pixels. One image was rewarded (CS+), the other not (CS-).   
They additionally had tactile cues provided by smooth vs. rough sandpaper attached to the side-arms of the maze. When choosing the rough sandpaper path in combination with CS+, they would receive a larger reward compared to smooth & CS+. 

So, for those experiments, I would choose to create a behavioralProtocol with 2 stimulations. 
**Stimulation 1:**
stimulusApproach: tactile stimulation
stimulusType: something along the lines of "texturalStimulus"  
**Stimulation 2:**
stimulusApproach: visual stimulation
stimulusType: something along the lines of "blackAndWhiteDigitalImage"

Any opinions/thoughts, @lzehl, @eapapp?

lzehl:
@UlrikeS91 thanks for responding so quickly. @eapapp thanks for raising the issue :)

@eapapp you should not be able to select any Stimulation Approach / Stimulus Type at the moment in the KGE because they are not connected to any v4 schema. @UlrikeS91 this is a bug we have to fix.

For StimulationApproach things are clear I think. And for now I would suggest to also purely stick to those. 

For StimulusType: We need to discuss this carefully with the ontology group. I would not like to rush this now and end up with something too painful to maintain. Independent of that here some points:
[...]

PR #65, opened 2021-11-04
New StimulationApproach/StimulusType instances

UlrikeS91:
related to #64 suggested by @eapapp

This is still a draft, so that we can add some additional stimulusType instances as well. 

So far, I added all stimulus modalities collected on InterLex as stimulationApproaches: https://neuinfo.org/interlex/view/ilx_0111059#children. This was anyway the plan. One of those is, of course, "visual stimulation" 😉

### DIFF
--- a/instances/stimulationApproach/auditoryStimulation.jsonld
+++ b/instances/stimulationApproach/auditoryStimulation.jsonld
@@ -9,7 +9,7 @@
   "interlexIdentifier": "http://uri.interlex.org/ilx_0101001",
   "knowledgeSpaceLink": null,
   "name": "auditory stimulation",
-  "preferredOntologyIdentifier": "http://www.cogpo.org/ontologies/CogPOver1.owl#COGPO_00125http://www.cogpo.org/ontologies/CogPOver1.owl#COGPO_00125",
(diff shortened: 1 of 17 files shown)


lzehl (approved):

lzehl:
@UlrikeS91 thanks :+1: is this ready to go then? you still have it as a draft

lzehl:
@UlrikeS91 I'm a bit confused...  and double checked now... We do not distinguish between stimulationApproach and stimulationType in any v4 schema (they were introduced for ephys which is not part of v4)... We only have Stimulation in BehavioralProtocol and Protocol. If we want to have them in we need to change the schemas as well.

UlrikeS91:
> @UlrikeS91 I'm a bit confused... and double checked now... We do not distinguish between stimulationApproach and stimulationType in any v4 schema (they were introduced for ephys which is not part of v4)... We only have Stimulation in BehavioralProtocol and Protocol. If we want to have them in we need to change the schemas as well.

Some confusion as stated in the related issue, so ignore this comment. This PR remains a draft until I've added the stimulusTypes needed for dataset that was described in the issue.

lzehl:
Thanks for keeping an eye on things if I'm getting confused :sweat_smile:

lzehl (approved):

PR #67, opened 2021-11-05
Age category updates

UlrikeS91:
Based on an offline discussion with @lzehl and @ingvildeb, we made some changes to the existing age catgeories:

1. "Post-juvenile adult" was removed (basically because it is a synonym for "adult")
2. "Adolescent" was added, defined as the stage right before and after puberty.
3. "Young adult" synonyms were updated ("adolescent" & "teenage" removed, "early adult (stage)" added)

This overview should put all age categories in context:
image


instances/ageCategory/adolescent.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/ageCategory/adolescent",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/AgeCategory",
+  "definition": "'Adolescent' categorizes the life cycle stage of an animal or human that starts right before puberty and ends right after puberty.",
```
lzehl: "adolescent" is a synonym for puberty. I would redefine this: "'Adolescent' categorizes a transitional life cycle stage of growth and development between childhood and adulthood, often described as 'puberty'". @ingvildeb: what do you think?

instances/ageCategory/adolescent.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "adolescent",
+  "preferredOntologyIdentifier": null,
+  "synonym": null
```
lzehl: "puberty"

lzehl (changes requested):

instances/ageCategory/adolescent.jsonld
ingvildeb: Sounds good to me!

lzehl (approved):

Issue #55, opened 2021-09-07
reassessing / extending TissueSampleType instances


lzehl:
list from @tgbugs on sample_types:
1. 'whole organism' (this we don't need, because this is a subject in our case)
2. 'whole organ' (is already in instances)
3. 'fluid specimen' (suggestion: integrate as "body fluid" ? )
4. 'tissue' (not sure what is meant here; we would need somthing like "tissue block"; cf. discussion below)
5. 'nerve' (could be integrated as is)
6. 'slice'  (is already integrated as "tissue slice")
7. 'section'  (see discussion points below)
8. 'cryosection' (I would not integrate this; mixes extraction / fixation procedure with tissue sample type)
9. 'cell' (suggestion: integrate as "single cell" ? )
10. 'nucleus' (I would not integrate this for now)
11. 'nucleic acid'  (I would not integrate this for now)
12. 'slide'  (see discussion points below)
13. 'whole mount' (see discussion points below)
14. 'cell suspension'  (could be integrated as is)
15. 'single cell suspension'  (not sure what the difference is to 'cell')

In openMINDS we registered so far:
1. 'biopsy sample' (cf. the above, I would keep this one)
2. 'brain hemisphere' (I would keep this one or is there a more general term, e.g. 'organ hemisphere' we would prefer?)
3. 'cell culture' (we discussed to replaces this with 'cell line' ?)
4. 'tissue slice' (cf. the above, I would keep this one)
5. 'whole organ' (cf. the above, I would keep this one)

**'slice' vs 'slide' vs 'section' vs 'whole mount' vs 'tissue block'**  
- **'slice'**: a tissue slice is for me just an unprocessed thin slice from a larger sample; does typically require specific sectioning / slicing devices
- **'slide'**:  a tissue slide is for me a tissue slice mounted on a object plate for, e.g., microscopy (can but does not have to be fixated and stained)
- **'section'**: is for me a synonym for slice
- **'whole mount'**: searched for it and mostly found "entire organism or structure which can be studied under microscope without further slicing" [so for me that would mean 'tissue block' is to 'whole mount' what 'slice' is to 'slide']
- **'tissue block'**: a tissue block measures for me a few centimeters or more; does not necessarily require specific devices (but a knife / scalpel)

@UlrikeS91 what do you think?
@tgbugs your thoughts??

UlrikeS91:
Sorry, this slipped through somehow...

Let's see:
[...]

PR #74, opened 2021-11-11
New and updated TissueSampleTypes

UlrikeS91:
#55


instances/tissueSampleType/biopsySample.jsonld
```diff
@@ -4,11 +4,11 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/tissueSampleType/biopsySample",
   "@type": "https://openminds.ebrains.eu/controlledTerms/TissueSampleType",
-  "definition": null,
+  "definition": "A sample of tissue or cell that was excised from a living or deceased multicellular organism body.",
```
lzehl: is that the definition from interlex? "or cell" seems weird to me... I would suggest:
"Typically a very small sample of tissue that was excised from a living or deceased multicellular organism body."

instances/tissueSampleType/hemisphere.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/tissueSampleType/hemisphere",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/TissueSampleType",
+  "definition": "A hemisphere that was excised from a bilateral organ tissue sample (e.g., a brain) from a living or deceased multicellular organism body.",
```
lzehl: Suggestion: "One of the symmetric halfs of an excised bilateral organ tissue sample ..."

instances/tissueSampleType/homogeneousCellPopulation.jsonld
```diff
@@ -2,13 +2,13 @@
...
-  "definition": "Cells taken from a living organism and grown under controlled conditions (in culture).",
+  "definition": "A sample of multiple cells/a population of cells that are of the same cell type.",
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
-  "name": "cell culture",
-  "preferredOntologyIdentifier": "http://purl.obolibrary.org/obo/BTO_0000214",
+  "name": "homogeneous cell population",
+  "preferredOntologyIdentifier": null,
   "synonym": null
```
lzehl, line 13: maybe here you should mention: cell line ?

instances/tissueSampleType/nerve.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/tissueSampleType/nerve",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/TissueSampleType",
+  "definition": "A nerve sample (i.e. a whole nerve or a part of a nerve) from a living or deceased multicellular organism body.",
```
lzehl, line 7: suggestion "... from a neuron of a ..."

lzehl (changes requested):

instances/tissueSampleType/biopsySample.jsonld
UlrikeS91: Yeah, it's from InterLex (modifed though). This is what they state: "A specimen collection that obtains a sample of tissue or cell from a living multicellular organism body for diagnostic purposes by means intended to be minimally invasive."

instances/tissueSampleType/homogeneousCellPopulation.jsonld
UlrikeS91: I wasn't sure about that. With broadening the name to homogeneous cell population, I don't think "cell line" would be a synonym anymore, but rather an example of such a population. What do you think?

instances/tissueSampleType/nerve.jsonld
UlrikeS91: I think we have different definitions for a nerve...

Merriam-Webster defines a nerve as "any of the filamentous bands of nervous tissue that connect parts of the nervous system with the other organs, conduct nerve impulses, and are made up of axons and dendrites together with protective and supportive structures". 

To me that's more than just a neuron, isn't it?

instances/tissueSampleType/biopsySample.jsonld
lzehl: Then use a mix:
 "Typically a very small sample of tissue that was excised from a living or deceased multicellular organism body for diagnostic purposes by means intended to be minimally invasive/destructive."

"destructive I've added for the deceased organisms... not sure if invasive makes sense there

instances/tissueSampleType/homogeneousCellPopulation.jsonld
lzehl: not sure. leave it out

instances/tissueSampleType/nerve.jsonld
lzehl: no ... I had the same definition in mind (should have written neurons though)... but leave it as it is

lzehl (approved):

PR #80, opened 2021-11-15
Add units core-hour and wafer-hour

apdavison:
Needed for resource usage provenance metadata (openMINDS/computation)

### DIFF
--- a/instances/unitOfMeasurement/coreHour.jsonld
+++ b/instances/unitOfMeasurement/coreHour.jsonld
@@ -1,0 +1,14 @@
+{
+    "@context": {
+      "@vocab": "https://openminds.ebrains.eu/vocab/"
+    },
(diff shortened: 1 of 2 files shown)


lzehl:
@apdavison do you need a review for this PR?

apdavison:
@lzehl yes please

lzehl (commented):
@apdavison thank you for this PR (:) quite important for computation, sorry we forgot about those...)

I have only one question: are there no abbreviations/symbols for core-hour and wafer-hour you could add as synoyms (cf. PR #61)?

apdavison:
since these are not very commonly used or familiar units, I think it's safest not to abbreviate them. "core-h" would be possible, but I think it would confuse most people.

lzehl:
@apdavison okay. then let's merge them, but we need to find a solution for #61 as well. could you have a look at that one as well?

PR #87, opened 2021-11-17
New breedingType instance

UlrikeS91:


lzehl (commented):
just for us as a reminder: we should add definitions for those very soon.

PR #88, opened 2021-11-18
Service instances

UlrikeS91:
EBRAINS Atlas Services was just a placeholder. Discussed offline with Lyuba to remove and add:
- siibra-explorer
- LocaliZoom
(not a complete list, but the two will need now)


instances/service/localiZoom.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/service/localiZoom",
```
lzehl: if it's a dedicated name that is capitalize you should capitalize it here as well

lzehl (approved):

lzehl (approved):

PR #96, opened 2021-12-01
Update and rename EthicsAssessments

lzehl:


lzehl:
@UlrikeS91 as announced (for the EBRAINS curation team) the changes for the ethics assessments. Could you review? Note: Your concern about "not required" (and your suggestion to change ending to "... most invertebrates.") should be addressed later and modified accordingly based on the feedback of the ethics team.

UlrikeS91 (approved):

Issue #101, opened 2021-12-09
potential controlled terms additions


lzehl:
How about these new TissueSampleAttributes (?):
+ sliced
+ untreated
+ mounted
+ fixated

And this new TissueSampleType (?):
+ tissue slice series

And this new SubjectAttribute (?):
+ comatose (OR unconscious)

UlrikeS91:
> How about these new TissueSampleAttributes (?):
> 
> * sliced
> * untreated
> * mounted
> * fixated

- sliced: I would say no. We have TissueSampleType "tissue slice" and I cannot come up with a use case were any other "TissueSampleType" would make sense in combination with sliced as "TissueSampleAttribute", maybe "nerve" + "sliced". But then I would rather introduce a new TSType...
- untreated: ok
- mounted: ok, maybe add "free floating" or similar in parallel?
- fixated: ok

On a general note: What is our perspective towards these atributes? As many details as possible or only crucial information? I'm asking because of terms like "mounted". It absolutely is a TSAttribute, but when would this information ever be crucial? It is procedural information which could also be captured in the Protocol/ProtocolExecutions. But then again all these attributes probably fall under this category, right?   
 
> And this new TissueSampleType (?):
> 
> * tissue slice series

I'm leaning towards no. TissueSampleCollections give part of this information already - not all of it, though. Could you describe what use case you had in mind? I find "series" is a bit vague, or was that on purpose? In the broadest sense "series" just means that they have something in common. With my background, I would assume tissue slices that were collected sequentially with equal spacing between the collected slices. If this is was you were aiming for, we may want to change the name slightly and it will need a good definition. 
Question is still if we want to go into this level of detail? So far, we kept TSTypes it quite high-level. I'm not against more detail, just putting the question out there ;)   


> And this new SubjectAttribute (?):
> 
> * comatose (OR unconscious)

I think "unconscious" might be misused, and there are still huge debates about what can be considered "(un)conscious". If "comatose" is related to a specific use case, where a patient actually is in a coma, we would colloquially use "unconscious" because we can't say whether or not that person is still self-aware, but maybe they are just "unresponsive" and unable to show that they are self-aware (use-case: locked-in syndrome). 
We also have "anaesthetized" which is more descriptive of the state the subject is in. We would assume that they would be unconscious then, but we don't know for certain. 
Long story short: I would add "comatose". It's more descriptive to the state the subject is in and less loaded with assumptions - similarly to "anaesthetized".

[...]

PR #103, opened 2021-12-13
New attributes

UlrikeS91:
related to #101

### DIFF
--- a/instances/tissueSampleAttribute/mounted.jsonld
+++ b/instances/tissueSampleAttribute/mounted.jsonld
@@ -4,7 +4,7 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/tissueSampleAttribute/mounted",
   "@type": "https://openminds.ebrains.eu/controlledTerms/TissueSampleAttribute",
-  "definition": "A specimen that has been preserved in a way that can be stored and handled easily for an extended period of time (e.g., on a microscope slide with mounting medium and covered with a coverslip).",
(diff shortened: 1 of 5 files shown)


instances/tissueSampleAttribute/mounted.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/tissueSampleAttribute/mounted",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/TissueSampleAttribute",
+  "definition": "A specimen that has been preserved in a way that can be stored and handled easily for an extended period of time (e.g., on a microscope slide with mounting medium and covered with a coverslip).",
```
lzehl: I don't like this one... although it often is includes preservation in histology, "mounted" does not necessarily mean it is preserved and can be handled on an extended period of time. You can also mount fresh specimen and then throw them away. Maybe that definition is more accurate (feel free to adapt though):

"A specimen is typically 'mounted' on a clear glass slide with aqueous or non-aqueous mounting media in order to be supported for further handling. Often 'mounted' specimen are additionally covered with a coverslip, especially for long term preservation."

lzehl (changes requested):

instances/tissueSampleAttribute/mounted.jsonld
UlrikeS91: Good suggestion. I shortened it a bit, though.

lzehl (approved):

PR #124, opened 2022-02-09
adding some definitions for selected techniques

lzehl:
@tgbugs please comment and correct:

An 'anatomical segmentation technique' fully annotates an anatomical image with regions of interest that might be useful for different types of image analysis.

An 'anatomical delineation technique' is a process in which a region of interest is segmented or parceled and then separated from the irrelevant portion of the image.

An 'anatomical parcellation technique' annotates substructures of an anatomical image with region(s) of interest that might be useful for different types of image analysis.


lzehl:
@tgbugs could you have a look if these definitions fit the terms?

tgbugs:
Results of discussion: merge all three of these into a single term, delineation technique, make parcellation and segmentation synonyms, add deterministic vs probabilistic. Note that some cleanup will need to be done inside ebrains as a result.

The basic reasoning is that segmentation and parcellation are used inconsistently between communities so it is better to make the properties such as completes and closedness explicit in cases where they are important (as with probabalistic/deterministic).

lzehl:
@tgbugs 'anatomical segmentation technique' and 'anatomical parcellation technique' instances are now deleted and the terms are stated as synonyms in 'anatomical delineation technique'. The definition of 'anatomical delineation technique' needs another review.

lzehl:
@tgbugs we should have a look at this one again this week so that we can resolve the conflicts and close this PR

instances/technique/anatomicalDelineationTechnique.jsonld
```diff
@@ -4,11 +4,15 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/anatomicalDelineationTechnique",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "An 'anatomical delineation technique' is a manual or automated process in which a structure or structures of interest in an anatomical image is/are annotated for further analysis or interpretation.",
```
tgbugs:
```suggestion
  "definition": "An anatomical delineation technique is any process that identifies the boundaries of an anatomical entity and applies those boundaries to specific data.",
```

lzehl (commented):
@aeidi89 I've integrated the definition provided by @tgbugs . Please double check and merge if this is okay now.

aeidi89 (approved):
Looks good

PR #125, opened 2022-02-10
Create computerTomography.jsonld

lzehl:


lzehl:
@tgbugs you might have this one. I did not check.

UlrikeS91:
> @tgbugs you might have this one. I did not check.

I found two that would fit but there is none that is in the specific methods/techniques ontology.
1. Computer Axial Tomography imaging protocol - http://uri.interlex.org/ilx_0102448 
2. Computer Axial Tomography imaging assay - http://uri.interlex.org/ilx_0102447

I think the second one could be used even though it's not specifically linked to the methods/techniques ontology (yet). The protocol portion fo the first one seems wrong.

lzehl:
@UlrikeS91 then lets merge first without interlex id, and add it later when we get confirmation.

UlrikeS91 (approved):

PR #116, opened 2022-01-14
Added definitions for some techniques

MaaikevS:

PR #127, opened 2022-02-10
new staining technique

lzehl:
@UlrikeS91 
+ "myelin staining" I would add this (similar logic like NisslStaining, correct?)
+ "thionin staining" add or cover differently (NisslStaining + used agent)???
+ "DAPi staining" add or cover differently???

### DIFF
--- a/instances/UBERONParcellation/cranialNerveII.jsonld
+++ b/instances/UBERONParcellation/cranialNerveII.jsonld
@@ -1,0 +1,23 @@
+{
+    "@context": {
+        "@vocab": "https://openminds.ebrains.eu/vocab/"
+    },
+    "@id": "https://openminds.ebrains.eu/instances/UBERONParcellation/cranialNerveII",
+    "@type": "https://openminds.ebrains.eu/controlledTerms/UBERONParcellation",
+    "definition": "The 'cranial nerve II' is comprised of retinal ganglion cell axons running posterior medially towards the optic chiasm, at which some of the axons cross the midline and after which the structure is termed the optic tract.",
+    "description": null,
+    "interlexIdentifier": "http://uri.interlex.org/base/ilx_0724691",
+    "knowledgeSpaceLink": "https://knowledge-space.org/wiki/UBERON:0000941#optic-nerve",
+    "name": "cranial nerve II",
+    "preferredOntologyIdentifier": "http://purl.obolibrary.org/obo/UBERON_0000941",
+    "synonym": [
+        "optic II",
+        "second cranial nerve",
+        "02 optic nerve",
+        "optic nerve [II]",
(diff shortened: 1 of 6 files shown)


UlrikeS91:
I checked InterLex - It has "myelin staining (protocol)" and I also checked the parent ("staining protocol" (http://uri.interlex.org/ilx_0111017)) which has:
- Golgi staining
- myelin staining
- Nissl staining
- silver staining (which has "Gallyas silver stain" and "Timm's stain")

I think that the 4 main categories can all be added as staining techniques. I'm not familiar with "Gallyas silver stain" but I do know that "Timm's stain" does more than just silver (other metals too) which is why I would argue to add this separately as well.

"Thionin staining" is not a technique but a protocol. The technique is "Nissl staining" where thionin is used as a reagent. So, I would not add this but rather add a rather standardized protocol in the EBRAINS KG that is title "thionin staining". 

"DAPi staining" should be a technique. It is a nuclear-specific dye that binds to DNA. There is no overarching technique that could describe this nicely and there is no overarching term used by the community. Theoretically, it could be grouped into something like "Nuclear DNA staining" but this does not exist. There are other stains that do similar things (like Hoechst stain), but everybody would refer to them by there specific names. 

Long story short, and while I'm at it, I will add:
- Golgi staining
- myelin staining
- Nissl staining
- silver staining
- Timm's staining
- DAPi staining
- Hoechst staining

I will not add:
- thionin staining: because it is Nissl staining with thionin as reagent
- Gallyas silver staining: mainly because I'm not familiar with it and are not aware of a use case that needs it, but this could be revisited whenever a use case comes up 

Any objections, @tgbugs, @lzehl or @majpuc?

lzehl:
@UlrikeS91 sounds good to me. big thank you for taking over :+1:

UlrikeS91:
Revision/Update to my list:

Golgi staining
--> is a silver staining technique, I don't see anything "special" with it, except for it following a specific protocol that is close to what Camillo Golgi has done when he invented it. I will not add it as a technique, should rather be a protocol with technique "silver staining"
 
myelin staining
--> @lzehl added this in this PR already, just updated the file with a definition

Nissl staining & silver staining
--> did not need to add them because they already existed but added information to silver staining (Nissl staining is being updated in #116)
 
Timm's staining
--> added, this is a combination of sulphide precipitation with silver staining (tissue metals to metal sulphide, metal sulphides catalyse silver ion reduction, built up of metallic grains)  

DAPi staining
Hoechst staining
--> both added

UlrikeS91:
@lzehl please review (can't actually add you as a reviewer because you made the PR).

lzehl:
@UlrikeS91 everything looks fine to me! Thanks a lot for taking over!

Issue #99, opened 2021-12-01
Microscopy techniques - new instance (and maybe a clean-up)


UlrikeS91:
@ingrre asked if we could add "confocal light-sheet microscopy" and I think that it would make sense to do so. @tgbugs I checked the method ontology but couldn't find it there. Would it make sense to add it? 

Additionally, I'm kind of wondering where we draw the line for the microscopy related terms? Below is a list of all microscopy term that we added so far:

- Brightfield microscopy
- Confocal microscopy
- Dual View Inverted Selective Plane Illumination microscopy
- Electron microscopy
- Epifluorescent microscopy
- Fluorescence microscopy
- Focused Ion Beam Scanning Electron microscopy
- Light microscopy
- Light Sheet Fluorescence microscopy
- Microscopy
- Polarized Light microscopy
- Scanning Electron microscopy
- Serial Section Transmission Electron microscopy
- Transmission Electron microscopy
- Two-Photon Excitation Fluorescence microscopy
- Widefield Fluorescence microscopy
- Widefield microscopy

The easy one first: "microscopy" needs to be removed. This is a duplication since it exists as an experimental approach (which obviously makes most sense anyway). 

Then, I think widefield fluorescence microscopy & epifluorescent microscopy might be synonym (see here: https://oni.bio/nanoimager/super-resolution-microscopy/epifluorescence-microscopy/).  

So, in short, my questions are:
1. Add "confocal light-sheet microscopy" - yes or no?
2. Remove "microscopy" (actual not a question)
3. "widefield fluorescence microscopy" same as "epifluorescent microscopy" - yes or no?
4. If 3. is answers with no, rename "epifluorescent microscopy" to "epifluorescence microscopy" to be consistent with the other fluorescence techniques?

@lzehl feel free to comment as well :)

lzehl:
@UlrikeS91 I agree to removing "microscopy" but maybe next year (if this is already in use in EBRAINS) I don't want to do another mapping right now for migration. But I will check how far this term is in use.

The rest I leave to @tgbugs

MaaikevS:
@UlrikeS91 @lzehl, "microscopy" is not in use so can be removed

lzehl:
@MaaikevS  would you prepare the PR for the removal of microscopy? I will delete it afterwards from the EBRAINS KG

MaaikevS:
[...]

PR #145, opened 2022-03-10
Delete microscopy.jsonld

MaaikevS:
Related to #99


lzehl:
@MaaikevS the respective EBRAINS KG instance is also deleted

PR #140, opened 2022-03-02
Added units of measurement

MaaikevS:
Related to Issue #134

PR #154, opened 2022-03-16
Add some commonly used electrical units

apdavison:
plus degree and radian


lzehl:
@apdavison can you compare the degree  / radian part please with the one from @MaaikevS PR #140 ? You two can decide who should delete / merge with the other. For degree: should be arcDegree (?) so that it does not collide with temprature degreeCelcius etc. OR move over to:
"degree" -> angle in deg (°)
"radian" -> angle in radian (rad)
"celcius" -> degree in celcius (°C)
"fahrenheit" -> degree in fahrenheit (°F) 
etc ???

apdavison:
To make things simple I've removed degree and radian from this PR.

lzehl (changes requested):
@apdavison thank you for this PR. Everything is good except that there is an empty line missing at the end of each JSONLD. Could you add this? (note: if you generate JSONs via Python you actively need to set this empty line. unfortunately Python does not follow the convention automatically.)

lzehl (approved):

Issue #153, opened 2022-03-16
correction of  openMINDS_controlledTerms/instances/service/LocaliZoom.jsonld 


Majpuc:
Hi, 
I discovered a misunderstanding in the definition of the Localizoom viewer as service link and would like to correct it. It is confusing but we have Localizoom as a viewer only, the one used as service link and then the Localizoom sw that is allowing annotations.
Here would be the correct text:
{
  "@context": {
    "@vocab": "https://openminds.ebrains.eu/vocab/"
  },
  "@id": "https://openminds.ebrains.eu/instances/service/LocaliZoom",
  "@type": "https://openminds.ebrains.eu/controlledTerms/Service",
  "definition": "Web application for viewing of series of high-resolution 2D images that have been anchored to reference atlases.",
  "description": "LocaliZoom allows the viewing and exploring of high-resolution images with superimposed atlas overlays.",
  "interlexIdentifier": null,
  "knowledgeSpaceLink": null,
  "name": "LocaliZoom",
  "preferredOntologyIdentifier": null,
  "synonym": null
}

lzehl:
@Majpuc is there in general a more lengthy description of LocaliZoom and documentation homepage links we should cover? It would be great to be a bit more detailed for the services are able to get these type of information. Could you support us here for LocaliZoom and the new one Multi-Image-OSd?

@UlrikeS91 FYI

UlrikeS91:
I guess @Majpuc came across this because of #152. I can fix it in parallel to adding the new service.

Majpuc:
yes, we have a Readthedocs for Localizoom:  [](https://localizoom.readthedocs.io/en/latest/index.htm)
https://localizoom.readthedocs.io/en/latest/index.htm
 The doc for Multi-Image OSd is coming soon.

Issue #152, opened 2022-03-16
Add new service instance - OSD viewer


UlrikeS91:
We are working on some additional service links for some dataset that will use a different tool than LocaliZoom, which is viewing any type of images without an atlas overlay or anything. @majpuc, could you please provide information about this tool so that we can add a new service instance for this tool, just like we have for LocaliZoom: https://github.com/HumanBrainProject/openMINDS_controlledTerms/blob/v1/instances/service/LocaliZoom.jsonld? 

Once I have that information, I can add the service that we will used in relation to the service links ☺️

Majpuc:
{
  "@context": {
    "@vocab": "https://openminds.ebrains.eu/vocab/"
  },
  "@id": "https://openminds.ebrains.eu/instances/service/Multi-Image-OSd",
  "@type": "https://openminds.ebrains.eu/controlledTerms/Service",
  "definition": "Web application for viewing of series of high-resolution 2D images.",
  "description": "Multi-Image-OSd allows the viewing and exploring of high-resolution images.  For more information see: https://github.com/Neural-Systems-at-UIO/Multi-Image-OSd",
  "interlexIdentifier": null,
  "knowledgeSpaceLink": null,
  "name": "Multi-Image-OSd",
  "preferredOntologyIdentifier": null,
  "synonym": null
}

MaaikevS:
I think we should push this if we want to move forward with the service links for this service. I would say the name probably has to be changed though.

PR #160, opened 2022-03-18
New services

UlrikeS91:
related to #152 & #153


MaaikevS (changes requested):
@UlrikeS91 the localizoom file is called "LocaliZoom.jsonl" instead of LocaliZoom.jsonld"

MaaikevS (approved):

PR #162, opened 2022-03-24
new techniques: high-throughput scanning & 4-points congruent sets alignment

lzehl:
+ high-throughput scanning
+ 4-points congruent sets alignment


instances/technique/highThroughputScanning.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/highThroughputScanning",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "'High-throughput scanning' is a technique for automatic creation of analog or digital images of a large numbers of samples.",
```
UlrikeS91: "...of large numbers of samples." OR "...of a large number of samples."

instances/technique/highThroughputScanning.jsonld
```diff
@@ -0,0 +1,16 @@
...
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "high-throughput scanning",
+  "preferredOntologyIdentifier": null,
+  "synonym": [
+    "high throughput scanning"
```
UlrikeS91, line 14: wouldn't abbreviation make sense to add as synonym - HTS? Or is this not commonly used? (not an expert, just asking)

UlrikeS91 (changes requested):

instances/technique/highThroughputScanning.jsonld
lzehl: HTS is typically used for high-throughput screening which is a different technique
in principle "high-throughput imaging" is more often used, but implies also image analysis, while the scanning would be analysis free. I'd like to distinguish this (as we do for microscopy techniques)

instances/technique/highThroughputScanning.jsonld
lzehl: corrected

UlrikeS91 (approved):

PR #165, opened 2022-03-30
Create AMPL.jsonld

jagru20:
Adding this programming language that is needed for https://github.com/bweyers/HBPVisCatalogue/issues/476 

@lzehl  @UlrikeS91  I hope this is the right way to suggest the term :D

### DIFF
--- a/instances/programmingLanguage/AMPL.jsonld
+++ b/instances/programmingLanguage/AMPL.jsonld
@@ -1,0 +1,14 @@
+{
+    "@context": {
+        "@vocab": "https://openminds.ebrains.eu/vocab/"
+      },
+    "@id": "https://openminds.ebrains.eu/controlledTerms/programmingLanguage/AMPL",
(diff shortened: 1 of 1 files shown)


lzehl (approved):
at one point we need to discuss wikidata under ontology identifier. this is in principle not correct I think. I would leave it for now though.

lzehl:
@UlrikeS91 any objections?

UlrikeS91:
> Adding this programming language that is needed for [bweyers/HBPVisCatalogue#476](https://github.com/bweyers/HBPVisCatalogue/issues/476)
> 
> @lzehl @UlrikeS91 I hope this is the right way to suggest the term :D

Yes, correct way to suggest new terms 😉 


> at one point we need to discuss wikidata under ontology identifier. this is in principle not correct I think. I would leave it for now though.

Agreed. Not ideal but for now better than leaving it blank. 


> @UlrikeS91 any objections?

Not an objection, but maybe add "algebraic modeling language" as synonym.

PR #169, opened 2022-04-05
Add instance for FTP repositories

skoehnen:
To support the Big Brain file repository it is necessary to support FTP data repositories as FileRepositoryType.

### DIFF
--- a/instances/fileRepositoryType/ftp.jsonld
+++ b/instances/fileRepositoryType/ftp.jsonld
@@ -4,7 +4,7 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/fileRepositoryType/ftp",
   "@type": "https://openminds.ebrains.eu/controlledTerms/FileRepositoryType",
-  "definition": "FTP, the file transfer protocol, is a standard internet communication protocol which allows the transfer of files between clients and a server.",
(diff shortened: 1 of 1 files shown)


instances/fileRepositoryType/ftp.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/fileRepositoryType/ftp",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/FileRepositoryType",
+  "definition": "FTP, the file transfer protocol, is a standard internet communication protocol which allows the transfer of files between clients and a server.",
```
lzehl: this should be the definition of an FTP repository not the FTP itself. something in the direction of (feel free to optimize): "An 'FTP repository' is located on a server that uses file transfer protocol (FTP), a standard ..."

lzehl (changes requested):
besides the definition (please modify correct my suggestion) this looks good

skoehnen:
Yes, that makes sense.
I will update definition.

instances/fileRepositoryType/ftp.jsonld
skoehnen: Yes, that makes sense.
I will update the definition.

lzehl (approved):

Issue #136, opened 2022-02-28
New deviceType suggestions


MaaikevS:
Would it be possible to extend the number of instances for "DeviceType"?

For example with:
- microscope
- vibratome
- microtome

UlrikeS91:
Sorry that nobody has reacted to your request. @lzehl, this was also discussed in relation to the ephys extension. Those can (and should be) added. The controlledTerms schema exists, and it has been added as a terminology, but the instances folder does not exist yet. 
@MaaikevS would you be able to prepare the instances in a new deviceType folder (here: https://github.com/HumanBrainProject/openMINDS_controlledTerms/tree/v1/instances)?

lzehl:
@MaaikevS @UlrikeS91 yes please go ahead and add instances for DeviceType. 
@Peyman-N please provide a list of DeviceTypes that you need (I assume for example "MRI scanner", or "PET scanner", or "Amplifier").

MaaikevS:
@UlrikeS91 @lzehl I was under the impression that deviceType already had a folder (there is the instance "magnetic resonance imaging scanner" for deviceType in the KGE). This is probably old then; the info can be converted into a new openMINDS instance and the instance can be removed from the KGE (it is not attached to anything yet)

I will start with the following device types:
- microscope
- vibratome
- microtome
- MRI scanner
- PET scanner
- Amplifier

@Peyman-N please let me know what else needs to be added.

MaaikevS:
Actually, I think PET is more a method than a device, since it is typically combined with CT (PET/CT) or MRI (PET/MRI). So perhaps we want to consider adding the combination instead of PET scanner. Thoughts?

lzehl:
as far as I researched there are also PET standalone scanners. But I agree we should add the more common ones as well PET/CT scanner (which is not purely a combination of two devices but one device able to perform both techniques as far as I know).

lzehl:
closed with #176

PR #176, opened 2022-04-08
Added instances device type

MaaikevS:
Related to #136


instances/deviceType/CTscanner.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
```
lzehl, line 2: modified the definition and added some synonyms

instances/deviceType/MRIscanner.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
```
lzehl, line 2: modified a bit the definition

instances/deviceType/amplifier.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
```
lzehl: modified a bit the definition and added some synonyms

instances/deviceType/microtome.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
```
lzehl, line 2: modified a bit the definition

instances/deviceType/vibratingMicrotome.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
```
lzehl, line 2: modified a bit the definition and changed the name to "vibrating microtome", because apparently "vibratome" and "compresstome" are not generic names but registered trademarks of different manufacturers

lzehl (approved):
@MaaikevS please double check my updates and tell me if we are good to merge from your side.

instances/deviceType/vibratingMicrotome.jsonld
MaaikevS: I would add vibratome as synonym, since this is very commonly used to refer to a vibrating microtome.
I prefer not to include "smallest slice thickness: ~ 50µm", since I cut brains at 40µm. You could change it to "typically cut at" if you want to include such a measure

lzehl: I agree to add "vibratome" as a synonym

"typically cut at" does not cover it since you can cut much larger sections with a vibratome.

-Let's leave the slice thickness limitations then out for "vibrating microtome" and "microtome"

lzehl: they can be stated in the actual registered devices (when they come)

lzehl (approved):

Issue #150, opened 2022-03-16
New cell types (based on term suggestions)


MaaikevS:
Based on the term suggestions, there is a number of cell tyes that we should consider adding:

aromatase expressing neuron
calbindin expressing neuron
calretinin expressing neuron
choline acetyltransferase expressing neuron
cortical neuron
cortical postmitotic cell
cortical progenitor cell
D2 receptor expressing neuron
excitatory neuron
hippocampal neuron
inhibitory neuron
layer 6 neuron
neuropeptide Y expressing neuron
nitric oxide synthase expressing neuron

@UlrikeS91 @lzehl Related to openMetadataInitiative/openMINDS_controlledTerms#149, let me know what naming convention you would like to use and I can prepare a PR.

lzehl:
@MaaikevS thanks for preparing this. There are a few that I would not register, because they are semantic combinations with brain regions and not represent necessarily their own class (but feel free to disagree, and discuss). I think we should avoid these where possible (which we did not in the past). @tgbugs your insight here? 

Here the once I would not register:
+ cortical neuron
+ cortical postmitotic cell (I would only register postmitotic cell)
+ cortical progenitor cell (I would only register progenitor cell)
+ hippocampal neuron
+ layer 6 neuron

For the rest: my impression is that the x expressing neurons are typically written as "x-expressing neuron", if the x is one word. Not sure if that is true. The decision of going with ... neuron now instead of ... cell is good I think.

All of those decisions we make now would require us to clean up what we already have in for cell types (in particular the combination of brain region with a general cell type). I think it would be worth it. 

@MaaikevS @UlrikeS91 @tgbugs what do you think?

MaaikevS:
Cell types that are now registered (and cleaned up in the KGE):

aromatase expressing neuron
calbindin expressing neuron
calretinin expressing neuron
choline acetyltransferase expressing neuron
postmitotic cell
progenitor cell
D2 receptor expressing neuron
excitatory neuron
inhibitory neuron
neuropeptide Y expressing neuron
nitric oxide synthase expressing neuron

Further discussion is needed to sort out cell types that refer to an anatomical region

lzehl:
this will be solved with https://github.com/HumanBrainProject/openMINDS_controlledTerms/issues/257

Issue #149, opened 2022-03-16
Reassessing cellType - neuron vs cell


MaaikevS:
@lzehl @UlrikeS91 We briefly talked about the naming convention of cell types offline. The thought was that we should use "neuron" if it can only be a neuron, otherwise "cell". It looks like there were some inconsistencies introduced in the past. Would this be something to clean up now (before adding more)?

This is the list we have at the moment:

astrocyte
**basket cell**
cerebellar   interneuron
**cerebellum   basket cell
**cerebellum   Golgi cell**
cerebellum   granule cell**
cerebellum   stellate neuron
**cholecystokinin   expressing cell**
cholinergic   interneuron
cholinergic   neuron
**cortical   basket cell**
cortical   interneuron
**D1 receptor   expressing cell**
dopaminergic   neuron
fast spiking   interneuron
**glial cell**
**granule cell**
hippocampus   CA1 pyramidal neuron
interneuron
**macroglial   cell
microglial   cell**
motor neuron
neocortex   layer 2/3 pyramidal neuron
neocortex   layer 5 tufted pyramidal neuron
neostriatum   cholinergic interneuron
neostriatum   direct pathway spiny neuron
neostriatum   indirect pathway spiny neuron
neuron
**parvalbumin   expressing cell**
**Purkinje   cell
pyramidal   cell**
sensory   neuron
**somatostatin   expressing cell**
spinal   interneuron
spiny neuron
stellate   neuron
striatal   interneuron
striatum   medium spiny neuron
**vascular   endothelial cell
vascular   smooth muscle cell**
**vasoactive-intestinal peptide expressing cell**

I would say, the following ones are the most relevant. If you want to keep it the way it is, then I feel that any new cell types should also have the name "cell" instead of "neuron":

cholecystokinin   expressing cell
D1 receptor   expressing cell
parvalbumin   expressing cell
somatostatin   expressing cell
vasoactive-intestinal peptide expressing cell

apdavison:
For the main label, I think we should go with the most commonly used name. For example, when I Google "Purkinje cell" (with the quotes) I get 955'000 results, but "Purkinje neuron" gives only 75'500 results. Of course we can and should put the synonyms in as well.

UlrikeS91:
Thanks, @MaaikevS for the very nice groundwork ☺️ This makes it a lot easier to discuss this topic.

[...]

PR #179, opened 2022-04-11
Added and changed cell type instances

MaaikevS:
Related to issues #149 and #150


lzehl (approved):
note: I've added also definitions for D1 and D2 Receptor Expressing Neurons

lzehl:
@MaaikevS after merging this PR we might need to do a cleanup in the EBRAINS KG as well (for all the ones that changed the atid now). Let me know if we should join forces.

PR #181, opened 2022-04-13
adding disorder of consciousness

lzehl:
I know that this is a disorder not disease, but I suggest we gather them for now in one terminology


MaaikevS (changes requested):
change "a subjects consciousness" to "a subject's consciousness" (possessive)

MaaikevS (approved):

Issue #185, opened 2022-04-13
New molecular entities


MaaikevS:
Here is a list of molecular entities that would be good to add. Naming convention (abbreviation vs full name) needs to be decided on.

I would like to propose to do this in stages and prioritise the more pressing ones first. Please feel free to indicate which ones need prioritisation.

name | synonym | interlex | preferredOntology | KnowledgeSpace
-- | -- | -- | -- | --
4-DAMP methobromide |   | http://uri.interlex.org/base/ilx_0686819 | http://id.nlm.nih.gov/mesh/2018/M0125168, https://ncimeta.nci.nih.gov/ncimbrowser/ConceptReport.jsp?dictionary=NCI%20Metathesaurus&code=C0096896 |  
acetylcholine | Ach | http://uri.interlex.org/ilx_0100240 | http://uri.neuinfo.org/nif/nifstd/sao185580330 |  
AgRP | agouti-related protein agouti-related peptide AGRP ART AGRT |   | http://purl.obolibrary.org/obo/PR_000003846 | https://knowledge-space.org/wiki/PR:000003846#agouti-related-protein
APV | (2R)-amino-5-phosphonovaleric acid AP5 | http://uri.neuinfo.org/nif/nifstd/nlx_94485 |   |  
Cav2.1 channel | P/Q voltage-dependent calcium channel |   |   |  
CGP 54626 |   | http://uri.interlex.org/ilx_0439399 | http://id.nlm.nih.gov/mesh/2018/M0279326 |  
cholecystokinin | CCK | http://uri.interlex.org/ilx_0102124 | http://uri.neuinfo.org/nif/nifstd/nifext_5068 |  
Choline Acetyltransferase | ChAT Choline acetylase | http://uri.interlex.org/base/ilx_0102129 | http://uri.neuinfo.org/nif/nifstd/sao722953401 |  
CNQX | cyanquixaline (6-cyano-7-nitroquinoxaline-2,3-dione) | http://uri.neuinfo.org/nif/nifstd/nlx_14886 |   |  
DAPi | 4,6-diamidino-2-phenylindole |   |   |  
delta opioid receptor | delta DOR | http://uri.neuinfo.org/nif/nifstd/nifext_6921 |   |  
DNQX | 6,7-dinitroquinoxaline-2,3-dione | http://uri.neuinfo.org/nif/nifstd/nlx_76042 |   |  
[...]

PR #188, opened 2022-04-13
add molecular entity instances

MaaikevS:
Made a start on implementing molecular entities with both an interlex and preferred ontology listed (issue #185).

Feel free to start reviewing!

### DIFF
--- a/instances/molecularEntity/dopamineTransporter.jsonld
+++ b/instances/molecularEntity/dopamineTransporter.jsonld
@@ -12,7 +12,6 @@
   "preferredOntologyIdentifier": "http://purl.obolibrary.org/obo/PR_000015188",
   "synonym": [
     "DAT",
-    "dopamine active transporter",
-    "SLC6A3"
+    "dopamine active transporter"
     ]
 }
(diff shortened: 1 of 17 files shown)


instances/molecularEntity/dopamineTransporter.jsonld
```diff
@@ -0,0 +1,18 @@
+{
+  "@context": {
```
lzehl, line 2: removed the SLC6A3 (this is a gene with the instruction for building the transporter, not the transporter itself)

instances/molecularEntity/flumazenil.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+      "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/molecularEntity/flumazenil",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/MolecularEntity",
+  "definition": "Flumazenil is a selective GABAA receptor antagonist",
```
lzehl: added the binding site (which is BZ one not the GABA one)

lzehl (approved):
@MaaikevS thank you for this PR. I've made some minor changes the rest looks okay to me.

Just to put it writing: On the long run I would like to go over the definitions and synonyms again for consistency and completion. What we want to do here we need to discuss first.

Issue #189, opened 2022-04-14
Request for new technique


szafarnia:
Dear all,

Would it be possible to add „Support Vector Regression” as a new technique?

Many thanks,
Sara

skoehnen:
Support Vector Regression is now available.

PR #190, opened 2022-04-14
Create supportVectorRegressionAlgorithm.jsonld

lzehl:
Related to #189 

cf. also https://github.com/HumanBrainProject/openMINDS_controlledTerms/blob/v1/instances/technique/supportVectorMachineClassifier.jsonld

(corrections can be suggested for both instances)


skoehnen:
@lzehl I wrote the definition more "statistically correct" or do you think this is too much?

skoehnen:
I removed the Regression part from the SVM definition and renamed the SVR technique so the name matches with SVC.

Issue #202, opened 2022-04-20
New techniques (from MRTM model variants)


AnnaHilverling:
Hi @skoehnen and @lzehl,

As a result of our discussion, I would like to have the following measurement methods entered as techniques: 
- binding potential measurement
- relative delivery rate measurement
- clearance rate measurement

Many thanks for your help!

lzehl:
@AnnaHilverling the PR is prepared. But I would like to get some expert feedback for those terms.

lzehl:
@AnnaHilverling I've discussed with @tgbugs and we decided to do the following with the MRTM model variants within EBRAINS:
+ register "multilinear regression" as technique
+ register "Multilinear Reference Tissue Model (MRTM)" and variants as keywords
+ register "PET-derived binding potential" etc. as study target

tgbugs:
https://www.pmod.com/files/download/v31/doc/pkin/2316.htm

UlrikeS91:
> @AnnaHilverling I've discussed with @tgbugs and we decided to do the following with the MRTM model variants within EBRAINS:
> 
> * register "multilinear regression" as technique
> * register "Multilinear Reference Tissue Model (MRTM)" and variants as keywords
> * register "PET-derived binding potential" etc. as study target

I've made a PR for the multilinear regression technique. The rest does not need to be handled here. PET-derived binding potential can be added as term suggestion for now. 

@lzehl, I do not now what the status for your draft PR #203 is, but to me it looks like this should be move somewhere else (e.g. MeasuredQuantity). The 3 terms form the PR do not seem suitable techniques.

PR #203, opened 2022-04-20
adding some pharmacokinetic techniques

lzehl:
related to #202


lzehl:
@tgbugs could you have a look at those? we are trying to get some feedback from PET experts but I think these methods (measurement techniques) are also used in other areas.

lzehl:
@AnnaHilverling I've discussed with @tgbugs and we decided to do the following with the MRTM model variants within EBRAINS:
+ register "multilinear regression" as technique
+ register "Multilinear Reference Tissue Model (MRTM)" and variants as keywords
+ register "PET-derived binding potential" etc. as study target

lzehl:
based on second discussions in #202 I'm closing this PR

Issue #205, opened 2022-04-20
UBERONparcellation


szafarnia:
Dear all,

Could you please add „intraparietal sulcus“ as UBERONparcellations? 
As Information: I found it in the ontology search: https://www.ebi.ac.uk/ols/ontologies/uberon/terms?iri=http%3A%2F%2Fpurl.obolibrary.org%2Fobo%2FUBERON_0002913

Many thanks in advance.

Best,
Sara

UlrikeS91:
Thanks for adding the ontology link, @szafarnia! I've created the instance. If either @MaaikevS or @lzehl reviews (and merges) it today, you can start using it tomorrow in the KGE ☺️

MaaikevS:
Approved and added!

PR #207, opened 2022-04-20
Create intraparietalSulcus.jsonld

UlrikeS91:
#205


MaaikevS:
@UlrikeS91 I will review this

MaaikevS (commented):
I would add the abbreviation IPS in brackets in the first sentence of the description. Now the term IPS comes out of nowhere

UlrikeS91:
> I would add the abbreviation IPS in brackets in the first sentence of the description. Now the term IPS comes out of nowhere

True. That's a copy-paste error. I missed that one, but I will rather add the abbreviation as you suggested (instead of writing it out in the other sentence.

MaaikevS (approved):

PR #211, opened 2022-04-22
Create interpolation.jsonld

lzehl:


skoehnen (approved):
The definition looks good to me.
But we should be careful when we use this technique, since it is quite broad.

skoehnen:
@lzehl Should we start to add some more specific interpolation methods?
From the top of my head I could add:

- linear, cosine, cubic, hermite and their (3D equivalents if we want to be that specific)
- Bi- and Tri-linear
- Different fitting functions
- Nearest neighbor

And I bet there are a lot more ;)

I keep this PR open for now, but maybe we should move this to a issue.

lzehl:
@skoehnen very good idea, but let's move this to another PR. you don't have to create and issue. would you take this PR over?

skoehnen:
> @skoehnen very good idea, but let's move this to another PR. you don't have to create and issue. would you take this PR over?

Yes, I will open a new PR with some more specific methods.

PR #213, opened 2022-04-29
Update microscopy techniques

MaaikevS:
@lzehl @UlrikeS91 I made a start on Issue #99


instances/technique/confocalMicroscopy.jsonld
```diff
@@ -4,11 +4,13 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/confocalMicroscopy",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Confocal microscopy is a specialized form of standard fluorescence microscopy that uses particular optical components to generate high-resolution images of material stained with fluorescent probes.",
```
UlrikeS91: Not 100% happy with the definition. For most people, there seems to be a big focus on the reduction/elimination of out-of-focus glare which speaks more to an increase contrast compared to widefield. Would be nice to include it. Maybe we could write:

"Confocal microscopy is a specialized fluorescence microscopy technique where laser light is focused onto a defined spot at a specific depth within a fluorescent sample to eliminate out-of-focus glare, and increase resolution and contrast in the micrographs."

instances/technique/electronMicroscopy.jsonld
```diff
@@ -4,11 +4,11 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/electronMicroscopy",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Electron microscopy is an electron-optical instrument in which a beam of electrons is used to produce an enlarged image of a minute object.",
```
UlrikeS91: microscopy is not an instrument

Easiest fix is to replace instrument by technique:
"Electron microscopy is an electron-optical technique in which a beam of electrons is used to produce an enlarged image of a minute object."

instances/technique/fluorescenceMicroscopy.jsonld
```diff
@@ -4,11 +4,11 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/fluorescenceMicroscopy",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Fluorescence microscopy comprises any type of microscopy where the specimen can be made to fluoresce (emit energy as visible light) by illuminating it with light of specific wavelengths (fluorophores).",
```
UlrikeS91: "... (emit energy as visible light) ..." isn't 100% correct; could we remove "visible"? The most accurate would be "electromagnetic radiation" which is overkill. It is typically visible light (as in light a human eye can detect), but it can also be light that is not within a human's visible spectrum. 

Also, could we remove "(fluorophores)"? It's true, but I think it should fine without it.

instances/technique/lightMicroscopy.jsonld
```diff
@@ -4,11 +4,14 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/lightMicroscopy",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Light microscopy, also referred to as a optical microscopy, comprises any type of microscopy technique that uses visible light to generate magnified images of small objects.",
```
UlrikeS91: For electron microscopy, you used "... to produce an enlarged image of a minute object." 

I find "...to generate magnified images of small objects." more understandable. Would it make sense to adjust the definition for electron microscopy accordingly?

instances/technique/lightSheetFluorescenceMicroscopy.jsonld
```diff
@@ -4,11 +4,13 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/lightSheetFluorescenceMicroscopy",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Lightsheet fluorescence microscopy is an optical microscopy technique that uses a thin sheet of light to excite only fluorophores within the focal volume.",
```
UlrikeS91: This might just be me being confused by all the different microscopy techniques, but above "optical microscopy" is the synonym for light microscopy. I've always associated light microscopy as the counterpart to fluorescence microscopy but this might be wrong? 

To me, fluorescence microscopy does not need to use visible light but light of a wavelength that can excite a fluorescent sample which then emits light that can be detected by the microscope and turned into an image that a human eye can perceive. Must the emission range be within the visual spectrum of a human? The excitation wavelength surely doesn't. 

I guess what I'm asking is:
Is "light microscopy" any microscopy technique using light (any light - within or outside of the visual spectrum of a human)?
If "yes", then I guess that is "fluorescence microscopy" a light microscopy technique.
If "no" and the definition must include "visible light" (only light within the visual spectrum of a human), then would "fluorescence microscopy" still be a light microscopy technique?

UlrikeS91 (changes requested):

instances/technique/electronMicroscopy.jsonld
tgbugs: Where is the `-optical` coming from here? Seems confusing?

instances/technique/fluorescenceMicroscopy.jsonld
tgbugs: You don't always have to excite.

instances/technique/electronMicroscopy.jsonld
tgbugs: Maybe: "Electron microscopy is a microscopy technique that uses differential scattering and or deflection of electrons passing through or bouncing off a sample to generate contrast."

instances/technique/electronTomography.jsonld
```diff
@@ -4,11 +4,13 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/electronTomography",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Electron tomography is a tomographic technique for obtaining 3-dimensional images with transmission electron microscopy.",
```
tgbugs: Electron tomography is a microscopy technique that take a series of images of a thick sample at different angles (tilts) so that tomography can be applied to increase the resolution of the ticker sample.

instances/technique/lightMicroscopy.jsonld
UlrikeS91: Remove the "a" before optical microscopy - "Light microscopy, also referred to as optical microscopy, ..."

instances/technique/lightSheetFluorescenceMicroscopy.jsonld
```diff
@@ -4,11 +4,13 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/lightSheetFluorescenceMicroscopy",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "Lightsheet fluorescence microscopy is an fluorescence microscopy technique that uses a thin sheet of light to excite only fluorophores within the plane of illumination.",
```
UlrikeS91: "a fluorescence" not "an fl." - "Lightsheet fluorescence microscopy is a fluorescence microscopy..."

instances/technique/widefieldMicroscopy.jsonld
```diff
@@ -6,9 +6,9 @@
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
   "definition": "'Widefield microscopy' comprises all microscopy techniques in which the entire sample is permanently exposed to a light source.",
```
UlrikeS91: Maybe remove the ' around widefield microscopy for consistency purposes (since they are not used in any of the other definitions? - "Widefiled microscopy comrpises all microscopy ..."

MaaikevS:
@UlrikeS91 I incorporated all the changes discussed online and offline.

UlrikeS91 (approved):

UlrikeS91:
Thank you @MaaikevS. I approved the PR now and will merge it. @tgbugs gave his feedback to all the definitions and the adjustments look good now.

PR #221, opened 2022-05-04
More microscopy updates

UlrikeS91:
related to #99


UlrikeS91:
I have worked my wait through the to-do list posted in the issue mentioned above. 

I additionally added "serial block face scanning electron microscopy" because I came across it while fixing all the other electron microscopy terms and had a definition that could be used. So, I may as well add it while it's fresh in my mind.

I looked in the requested "confocal light-sheet microscopy" and this seems to be wrong. The image below visualizes the 3 major illumination strategies - confocal, light-sheet and widefield - quite nicely. Light-sheet microscopy illuminates the sample from the side with a thin sheet of light (which increase the contrast and decreases the out-of-focus glare while increasing the scanning speed significantly compare to confocal microscopy). Confocal microscopy illuminates the sample from below through small pinholes (one or several as is the case for spinning disk confocal microscopy).   

image

I asked @ingrre to dig up the use case this technique was requested for, and this is the related publication: https://pubmed.ncbi.nlm.nih.gov/26158018/. They did create a new type of light-sheet microscope and referred to it as "custom-made confocal light-sheet microscope" (2.1. Laser Unit) and/or as "double-sided light-sheet microscope" (in the discussion). 
I'm not sure what to do with this, but the simplest solution might be to add one of these terms (or both) as keywords to the dataset (https://search.kg.ebrains.eu/live/minds/core/dataset/v1.0.0/dbd91bef-21ba-4922-b797-cf3a43e8cf50). They have not been very consequent in what they would like this new microscope to be called and it is unclear to me why this would be confocal (but that might just be caused by my limited knowledge). @tgbugs, @MaaikevS and @lzehl what do you think?

tgbugs:
LGTM!

tgbugs (approved):

PR #229, opened 2022-05-30
Update Contribution file

lzehl:
... to include a guideline for providing instances for controlled terms.

@MaaikevS & @UlrikeS91 please have a look if the edition is understandable. I'm open for suggestions. We should keep it short though.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,7 +40,7 @@ Non-backwards compatible changes are only accepted for version branches that do 
 
 For all contributions, make sure that you describe well what you plan to do / have done to speed up the evaluation / review process of your issue / pull request. The assigned member of the openMINDS development team will get back to you as soon as possible to discuss or give the blessing to your contribution. Note that, all contributing community members are welcome to provide comments on active issues and pull requests. 
 
-**Guidelines to contribute new controlled instances:** Terminology instances for any terminology all have the same basic schema against they are validated. Please use the following template to provide new controlled instance for an openMINDS controlledTerms terminology schema:
+**Guidelines to contribute new controlled instances to existing terminologies:** All controlled instances across all terminologies are validated against the same basic schema. Please use the following template to provide new controlled instance for an existing openMINDS controlledTerms terminology:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -40,6 +40,26 @@ Non-backwards compatible changes are only accepted for version branches that do
 
 For all contributions, make sure that you describe well what you plan to do / have done to speed up the evaluation / review process of your issue / pull request. The assigned member of the openMINDS development team will get back to you as soon as possible to discuss or give the blessing to your contribution. Note that, all contributing community members are welcome to provide comments on active issues and pull requests. 
 
+**Guidelines to contribute new controlled instances:** Terminology instances for any terminology all have the same basic schema against they are validated. Please use the following template to provide new controlled instance for an openMINDS controlledTerms terminology schema:
```
UlrikeS91: **Guidelines to contribute new controlled instances to existing terminologies:** All controlled instances across all terminologies are validated against the same basic schema. Please use the following template to provide new controlled instance for an existing openMINDS controlledTerms terminology:

CONTRIBUTING.md
```diff
@@ -40,6 +40,26 @@ Non-backwards compatible changes are only accepted for version branches that do
...
+  "preferredOntologyIdentifier": "Enter here the IRI of the preferred ontology identifier. Set value to ’null’ if not available.",
+  "interlexIdentifier": "Enter here the IRI of the interlex identifier (cf. [InterLex Project](https://neuinfo.org/interlex/dashboard)). Set value to ’null’ if not available.",
+  "knowledgeSpaceLink": "Enter here the IRI of the KnowledgeSpace term (cf. [KnowledgeSpace](https://knowledge-space.org/)). Set value to ’null’ if not available.",
+  "synonym": [
+    "Enter here all EXACT synonyms as a list. Set the whole value to ’null’ if not available."
+  ]
+}
+```
+
+Note that the optional properties are highly desired if they are available. In particular, please try to always provide a definition  for your new instance suggestion.
```
UlrikeS91: one space too many between "a definition" and "for your new"

UlrikeS91 (changes requested):
This looks good. The first part was a bit hard to read. Feel free to adjust my suggestion, too. 

One general thing: What you are describing here is the contribution of a single term to an existing terminology. I would point this out more clearly. The other option is contributing by add a new terminology (and at least one term), which would be different from what you describe here. 
We should add this at some point too, but it doesn't need to be right away.

CONTRIBUTING.md
lzehl: good suggestion. thanks. I'll copied it in

UlrikeS91 (approved):

PR #256, opened 2022-07-20
new technique: Golgi staining 

UlrikeS91:


instances/technique/golgiStaining.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/golgiStaining",
```
lzehl: this should be capitalized (.../GolgiStaining)since Golgi is a proper name.

instances/technique/golgiStaining.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Any of several methods for staining nerve cells, nerve fibers, and neuroglia in which fixed tissue is impregnated with silver nitrate and potassium dichromate resulting in the complete staining of some nerve cells while other cells are not stained at all.",
```
lzehl: I would change this a bit (even though this is the definition on interlex).

"'Golgi staining' includes several silver staining techniques in which fixed tissue is impregnated with silver nitrate and potassium dichromate resulting in the complete staining of some nerve cells while other cells are not stained at all." with referencing original interlex source and wikipedia

instances/technique/golgiStaining.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "description": null,
+  "interlexIdentifier": "http://uri.interlex.org/ilx_0104713",
+  "knowledgeSpaceLink": null,
+  "name": "Golgi staining",
+  "preferredOntologyIdentifier": "http://uri.neuinfo.org/nif/nifstd/birnlex_2243",
+  "synonym": null
```
lzehl: Golgi method
black reaction
?

lzehl (changes requested):

instances/technique/golgiStaining.jsonld
UlrikeS91: true, fixed

instances/technique/golgiStaining.jsonld
UlrikeS91: referenced interlex and not wikipedia, because I didn't use it (only interlex)

instances/technique/golgiStaining.jsonld
UlrikeS91: added

lzehl (approved):

PR #259, opened 2022-08-16
New molecular entities

UlrikeS91:
Rest of issue #185


lzehl:
@UlrikeS91 thanks a lot for this big PR and the work you already put into it.
I going over all terms and try to add missing information, optimize definitions, etc.
[done: 17/53]

One issue which is not easy to figure out is, that based on the terms you had to work with, the differentiation between subtypes of channels and receptors vs subunits of channels and receptors. I've tried to solve this but it is quite tricky.

First question that came up now in this context:
+ G protein-coupled inwardly-rectifying potassium channels -> renaming to Kir3 potassium channels ? (subtypes are registered under Kir3.1-4 in InterLex) [I would adapt my edition of the term depending on your feedback]

UlrikeS91:
@lzehl, thanks for reviewing and revising them!

> * G protein-coupled inwardly-rectifying potassium channels -> renaming to Kir3 potassium channels ? (subtypes are registered under Kir3.1-4 in InterLex) [I would adapt my edition of the term depending on your feedback]

I guess that is fine. I'm not really an expert to tell you for sure that those are the same in a way. If we lack any term in the end, we can still add more. Those will be correct, just maybe not the one that was needed. Let go for your suggestion and figure out the rest later 😉

lzehl:
@tgbugs and @aeidi89 I need help taking this one over; @UlrikeS91 thanks for making the initiating this PR for all these terms

lzehl:
@tgbugs @aeidi89 we need to get back to this one. I can't remember our conclusions of the discussion we already had on the topic.

UlrikeS91:
Tracked in the instances-repo. Needs to be restructured. Therefore I'm closing the PR.

Issue #212, opened 2022-04-26
Echo-planar imaging


AnnaHilverling:
Hi @UlrikeS91, @MaaikevS and @lzehl,

I have the question if echo-planar imaging (EPI) should be listed as a separate technique?
Actually EPI is an MRI pulse sequence and therefore a parameter of MRI. In this form I also found only echo planar pulse sequence via Interlex: https://scicrunch.org/scicrunch/interlex/view/ilx_0103659?searchTerm=echo-planar%20imaging

But how else could EPI be registered? Should it simply be specified as a keyword?

Thanks for your opinions in advance!

lzehl:
@tgbugs we should discuss this one

lzehl:
@AnnaHilverling: based on the discussion with @tgbugs we will register "echo planar pulse sequence" into techniques. 

For curation: register magnetic resonance imaging + the pulse sequence

Other pulse sequences to register:
+ T1 pulse sequence (T
+ T2 pulse sequence (T2 prepared)
+ etc

[look at: https://pubs.rsna.org/doi/10.1148/rg.262055063 and wikipedia MRI sequence page]

UlrikeS91:
What's the status here?

lzehl:
@UlrikeS91 was not done yet. The decision requires a potential clean up of other pulse sequences that were mixed with the acquisition technique (e.g., T1magneticResonanceImaging).

TODO:
1) registration of "echo planar pulse sequence"
2) registration of "T1 pulse sequence"
3) registration of "T2 pulse sequence"
4) removal of "T1-weighted magnetic resonance imaging", cleanup in EBRAINS (MRI + T1 pulse sequence technique)
5) removal of "T2-weighted magnetic resonance imaging", cleanup in EBRAINS (MRI + T1 pulse sequence technique)
6) screening the existing techniques for other cleanups (and correcting them)

lzehl:
Summary of what to define in #262 (@UlrikeS91 for you as support)

term name: **"T1 pulse sequence"**
definition: "In magnetic resonance imaging, a 'T1 pulse sequence' is a contrasting technique that allows the magnetization of the specimen or object to recover (spin-lattice relaxation) before measuring the magnetic resonance signal by changing the repetition time. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Receptive_field)]"
[...]

PR #262, opened 2022-08-19
MRI technique updates

UlrikeS91:
related to #212 

I started preparing the points from [this comment](https://github.com/HumanBrainProject/openMINDS_controlledTerms/issues/212#issuecomment-1189994659). I struggled with the definitions for the new sequence techniques, so any help would be appreciated @lzehl and @tgbugs.

We can only start the cleanup, when the new techniques can be used in the KG.


apdavison:
@Peyman-N do you have any comments here, given your background in MRI and your work on an openMINDS/MRI extension?

lzehl (changes requested):
@UlrikeS91 thanks for preparing this. 

I think we don't need to use "... technique" here. Just go with the "xxx sequence".
I'll prepare a summary in separate comment (in the related issue).

UlrikeS91:
Updated instances based on the comments in the issue #212

lzehl (approved):

Issue #227, opened 2022-05-19
new technique: retinotopic mapping


szafarnia:
Dear all,

Could you please add "retinotopic mapping" as new technique to KG?

Many thanks in advance.

Best,
Sara

lzehl:
@tgbugs @UlrikeS91 @MaaikevS could we discuss this in the ontology meeting? I would argue that this fits since we also support receptive field mapping.

tgbugs:
The two techniques are separate. Most simply, receptive fields can apply to more than the visual system. Receptive field mapping is usually to single cells where as retinotopic mapping is whole populations/swaths of cortex. See https://doi.org/10.1016/j.neuron.2016.07.015 for an example of a transition between the two levels.

PR #263, opened 2022-08-19
Receptive fields and retinotopic mapping

UlrikeS91:
related to #227 

I'm not sure my definitions are 100% accurate but this is what I could find out about the techniques. One major note is that RF mapping can be performed for other sensory areas as well - not only for visual RFs. 

@tgbugs and @lzehl

### DIFF
--- a/instances/technique/receptiveFieldMapping.jsonld
+++ b/instances/technique/receptiveFieldMapping.jsonld
@@ -4,7 +4,7 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/receptiveFieldMapping",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "'Receptive field mapping' is use to map out which portion of sensory space elicit neuronal responses when stimulated.",
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
--- a/instances/technique/receptiveFieldMapping.jsonld
+++ b/instances/technique/receptiveFieldMapping.jsonld
@@ -4,7 +4,7 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/receptiveFieldMapping",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": "'Receptive field mapping' is use to map out which portion of sensory space elicit neuronal responses when stimulated.",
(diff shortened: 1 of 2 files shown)


instances/technique/receptiveFieldMapping.jsonld
```diff
@@ -4,11 +4,11 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/receptiveFieldMapping",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "'Receptive field mapping' is use to map out which portion of sensory space elicit neuronal responses when stimulated.",
```
lzehl: **Alternative definition:**  "'Receptive field mapping' is a stimulation technique used to map out which neuron or neuron population preferably elicit a neuronal response to a specific sensory stimulus or sensory stimulus set." 

@tgbugs I think we need your additional feedback here for a nice definition.

instances/technique/retinotopicMapping.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/retinotopicMapping",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "'Retinotopic mapping' is a special type of receptive field mapping where the visual receptive field position is mapped in retinotopic coorinates in a brain region.",
```
lzehl: Better definition: "'Retinotopic mapping' is a stimulation technique in which neurons or neuron populations within the visual stream are mapped point-to-point to visual input of the retinal surface.

@tgbugs also here I think we need you feedback for a nice definition.

lzehl (changes requested):
@tgbugs we need your help with the definitions :)

lzehl:
@UlrikeS91 I've made some changes in the definitions. Please double check. If you are okay with the changes I suggest we merge this one. Corrections of the definitions by @tgbugs can be integrated later (if necessary).

UlrikeS91 (commented):
I did a very minor update to the definition of the RF mapping (a "to" was missing) and added "RF mapping" as synonym. 

I can't approve my own PR (therefore I only comment). But I think the changes are good and we can merge now.

lzehl (approved):

PR #270, opened 2022-08-29
Add DifferenceMeasure

apdavison:
This is needed initially for an extension to openMINDS/computation covering model validation tests, however it is also a starting point for planned future formalization of data analysis methods.


instances/differenceMeasure/kullbackLeiblerDivergence.jsonld
```diff
@@ -0,0 +1,16 @@
...
+    "@id": "https://openminds.ebrains.eu/instances/differenceMeasure/kullbackLeiblerDivergence",
+    "@type": "https://openminds.ebrains.eu/controlledTerms/DifferenceMeasure",
+    "definition": "A measure of how one probability distribution is different from a second, reference probability distribution.",
+    "description": null,
+    "interlexIdentifier": null,
+    "knowledgeSpaceLink": null,
+    "name": "Kullback-Leibler divergence",
+    "preferredOntologyIdentifier": null,
+    "synonym": [
+        "KS-statistic"
```
lzehl: KL-statistic ?

instances/differenceMeasure/meanSquaredError.jsonld
```diff
@@ -0,0 +1,14 @@
...
+    "@context": {
+      "@vocab": "https://openminds.ebrains.eu/vocab/"
+    },
+    "@id": "https://openminds.ebrains.eu/instances/differenceMeasure/meanSquaredError",
+    "@type": "https://openminds.ebrains.eu/controlledTerms/DifferenceMeasure",
+    "definition": "The mean squared difference between two series of values.",
+    "description": null,
+    "interlexIdentifier": null,
+    "knowledgeSpaceLink": null,
+    "name": "z-score",
```
lzehl: copy paste error: should be "mean squared error"
does this conflict / duplicate the TypeOfUncertainty term ?

instances/terminology/differenceMeasure.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@id": "https://openminds.ebrains.eu/instances/terminology/differenceMeasure",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Terminology",
+  "definition": "A measure of the difference between two things",
+  "description": "This may be a numerical or physical quantity, a set of categories, etc. Examples include 'mean squared error', 't-statistic', 'p-value'.",
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "measured type",
+  "preferredOntologyIdentifier": null,
+  "synonym": null
+}
```
lzehl: new line missing at the end of the JSON

lzehl (changes requested):
@apdavison aside the small copy/paste errors I found this seems okay. I just have one question in general: does this conflict / duplicate the TypeOfUncertainty terms (in particular the mean squared error)? should we merge them or can we more clearly differentiate them?

instances/differenceMeasure/kullbackLeiblerDivergence.jsonld
apdavison: fixed

instances/terminology/differenceMeasure.jsonld
apdavison: fixed

instances/differenceMeasure/meanSquaredError.jsonld
apdavison: fixed copy-paste error

apdavison:
there's clearly some overlap with TypeOfUncertainty - that seems more narrowly defined than DifferenceMeasure.

apdavison:
We could certainly merge them. What would be the downsides of that, as far as you can see?

lzehl:
@apdavison I don't see a real downside. potentially that some methods are not only meant for uncertainty measurements, but you have a better insight in how far this would be confusing in the QuantitativeValue uncertainties. What do you think?

lzehl:
Merging the TypeOfUncertainty into the DifferenceMeasure may require a KG cleanup / migration in EBRAINS, but that cannot be much and should be doable

apdavison:
Handling this properly would require either supporting a hierarchical structure for controlled terms (with types and sub-types), or providing an enum-type mechanism in the schema template language, where we could specify specific controlled term instances.

What about keeping them separate for now, and then next time a similar situation comes up we decide with Oli and perhaps Tom G. how to handle this in the general case?

lzehl:
@apdavison agreed.

apdavison:
ok to merge?

instances/terminology/differenceMeasure.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/terminology/differenceMeasure",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Terminology",
+  "definition": "A measure of the difference between two things",
+  "description": "This may be a numerical or physical quantity, a set of categories, etc. Examples include 'mean squared error', 't-statistic', 'p-value'.",
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "measured type",
```
lzehl: I overlooked this one... name is "difference measure"

lzehl (changes requested):

instances/terminology/differenceMeasure.jsonld
apdavison: done

lzehl (approved):

Issue #271, opened 2022-08-30
new technique: ELISA (Enzyme-linked Immunosorbent Assay)


szafarnia:
Dear all,

Could you please add „ELISA (Enzyme-linked Immunosorbent Assay)” as a technique to the KG?

Many thanks in advance.

Best,
Sara

lzehl:
@tgbugs and @UlrikeS91 what is our position on assays? do we register them (for now) under techniques if they are well established?

UlrikeS91:
ELISA is definitely well established. I think it should be fine to add this one as a technique even though it may not fall into this category in the classical sense since it is more like a protocol.

PR #277, opened 2022-09-01
Create enzymeLinkedImmunosorbentAssay.jsonld

lzehl:
related to  #271


UlrikeS91 (approved):
I just removed an extra space in the definition. Rest is fine. Will merge it.

Issue #283, opened 2022-09-02
new Technique: diagnostic assessment


lzehl:
for all the diagnostic checklists and other medical assessments 

@UlrikeS91 @tgbugs what do you think?

UlrikeS91:
Sounds good! I support this.

PR #285, opened 2022-09-02
Create psychologicalTesting.jsonld

lzehl:
related to #283 

based on some more research I assume the "technique" behind checklists like CBCL, etc is better described with "psychological testing"

### DIFF
--- a/instances/technique/psychologicalTesting.jsonld
+++ b/instances/technique/psychologicalTesting.jsonld
@@ -1,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/psychologicalTesting",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "'Psychological testing' is the administration of psychological tests typically by trained evaluators which also evaluate the person's responses according to carefully prescribed guidelines resulting in psychometric scores are thought to reflect individual or group differences in the construct the test purports to measure. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Psychological_testing)]",
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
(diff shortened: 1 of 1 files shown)


UlrikeS91:
> related to #283
> 
> based on some more research I assume the "technique" behind checklists like CBCL, etc is better described with "psychological testing"

Sounds good. Would this mean that there should also be a technique called "physical testing"?

instances/technique/psychologicalTesting.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/psychologicalTesting",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "'Psychological testing' is the administration of psychological tests typically by trained evaluators which also evaluate the person's responses according to carefully prescribed guidelines resulting in psychometric scores are thought to reflect individual or group differences in the construct the test purports to measure. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Psychological_testing)]",
```
UlrikeS91: The definition is a bit long and confusing. 

Definition: "'Psychological testing' is a psychometric measurement to evaluate a person's response to a psychological test according to carefully prescribed guidelines. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Psychological_testing)]"
 
The rest could be moved into the description.

UlrikeS91 (changes requested):

instances/technique/psychologicalTesting.jsonld
lzehl: I like it. Thanks. And I updated it. Let's do a more detailed description later.

UlrikeS91 (approved):

instances/technique/psychologicalTesting.jsonld
lzehl: but crashed the code... wait a moment

Issue #278, opened 2022-09-01
new UBERONparcellations: blood, arterial blood, venous blood


lzehl:
The following terms are currently not registered in our UBERONparcellations terminology:

blood
http://purl.obolibrary.org/obo/UBERON_0000178

arterial blood
http://purl.obolibrary.org/obo/UBERON_0013755

venous blood
http://purl.obolibrary.org/obo/UBERON_0013756

@ehennestad it would be great if you could do the respective PR for those terms. 
Here a summary of where the terms (as JSONLDs) need to go and how they should be registered:

**Where should they be added?** 
https://github.com/HumanBrainProject/openMINDS_controlledTerms/tree/v1/instances/UBERONParcellation

**How should they be added?**
EXAMPLE: https://github.com/HumanBrainProject/openMINDS_controlledTerms/blob/v1/instances/UBERONParcellation/barrelCortex.jsonld
The **definition** should always describe the anatomical organization of the term (phrase needs of course to be adapted a bit depending on the term): "'TERM' is a XXX of YYY. It is part of the ZZZ."
The **description** just takes whatever is additionally provided through the UBERON ontology or the InterLex entry (see below).
The **interlexIdentifier** can be found by searching the term on the InterLex portal: https://neuinfo.org/interlex/dashboard (Note: not all terms are already registered in InterLex)
The **knowledgeSpaceLink** can be found by searching the term on Knowledge Space: https://knowledge-space.org/search (under TERMS)
The **name** is always the same name as used in UBERON: https://www.ebi.ac.uk/ols/ontologies/uberon
The **preferredOntologyIdentifier** is always the UBERON ontology identifier.
The **synonym** is always the same list as given under "has exact synonym" in the UBERON ontology.

PR #286, opened 2022-09-02
Implement issue #278

ehennestad:
Add controlledTerms for blood, arterialBlood, venousBlood

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,0 @@
-.DS_Store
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -0,0 +1 @@
+.DS_Store
```
lzehl: you need to exclude this file from the commit.

lzehl (changes requested):
@ehennestad thank you so much for taking over this PR. Could you remove the gitignore file from the PR though? rest looks good :)

lzehl (approved):

Issue #289, opened 2022-09-05
new technique: magnetoencephalography


szafarnia:
Dear all,

Can “magnetoencephalography” be added as new technique to KG?

Best,
Sara

lzehl:
@szafarnia of course. I'm surprised we don't have that in yet.

PR #290, opened 2022-09-05
Create magnetoencephalography.jsonld

lzehl:
related to #289


UlrikeS91 (approved):
I've added the InterLex and pref. ontology ID. If that ok, we can merge. The rest looks fine to me 🙂

lzehl:
@UlrikeS91 yes fine with me (just for the record: we have to discuss the ontology IDs that are actually leading back to interlex with @tgbugs at one point)

Issue #276, opened 2022-08-31
new experimental approach: pharmacology


lzehl:
@tgbugs and @UlrikeS91 should we add this to exp approaches?

UlrikeS91:
I support this suggestion.

PR #291, opened 2022-09-06
Create pharmacology.jsonld

lzehl:
related to #276

### DIFF
--- a/instances/technique/pharmacology.jsonld
+++ b/instances/technique/pharmacology.jsonld
@@ -10,8 +10,5 @@
   "knowledgeSpaceLink": null,
   "name": "pharmacology",
   "preferredOntologyIdentifier": "http://edamontology.org/topic_0202",
-  "synonym": [
-    "computational pharmacology",
-    "pharmacoinformatics"
-  ]
+  "synonym": null
 }
--- a/instances/technique/pharmacology.jsonld
+++ b/instances/technique/pharmacology.jsonld
@@ -1,0 +1,17 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/pharmacology",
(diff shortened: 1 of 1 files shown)


instances/technique/pharmacology.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "'Pharmacology' is an experimental approach in which the composition, properties, functions, sources, synthesis and design of drugs (any artificial, natural, or endogenous molecule) and their biochemical or physiological effect (normal or abnormal) on a cell, tissue, organ, or organism are studied. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Pharmacology)]",
+  "description": null,
+  "interlexIdentifier": "http://uri.interlex.org/base/ilx_0108784",
+  "knowledgeSpaceLink": null,
+  "name": "pharmacology",
+  "preferredOntologyIdentifier": "http://edamontology.org/topic_0202",
+  "synonym": [
+    "computational pharmacology",
+    "pharmacoinformatics"
```
UlrikeS91: Those are narrow synonyms - remove.

UlrikeS91 (changes requested):

instances/technique/pharmacology.jsonld
lzehl: @UlrikeS91 I agree but the ontologies do not... should we anyway remove them?

lzehl: I just removed them. I had the same issue with them (and in interlex those are also classified as narrow, not exact)

lzehl: And actually the EDAM lists them twice (once under narrow and once under synonyms) ... @UlrikeS91 good to go from my side with synonyms set to null for us

UlrikeS91 (approved):

PR #293, opened 2022-09-08
Create checkerboardVisualStimulation.jsonld

lzehl:
missing from current stimulation technique set (natural image, grating etc)

### DIFF
--- a/instances/technique/checkerboardVisualStimulation.jsonld
+++ b/instances/technique/checkerboardVisualStimulation.jsonld
@@ -1,0 +1,19 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/checkerboardVisualStimulation",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Stimulation technique that uses a checkerboard as visual stimulus.",
+  "description": null,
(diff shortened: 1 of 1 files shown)


UlrikeS91 (approved):

UlrikeS91:
I was a bit too quick here! This should not have been merged into techniques. We have checker board stimuli, but under "stimulusType" where it belongs. Sorry about the extra work, but we should delete it again...

lzehl:
@UlrikeS91 no based on what we have already in techniques this has to go here (as well)

lzehl:
there is the point of stimulation VS stimulus

UlrikeS91:
Yes, this is why it should be added as a [stimulation](https://github.com/HumanBrainProject/openMINDS_core/blob/v4/schemas/research/stimulation.schema.tpl.json) which has a stimulation approach (visual stimulation) and a stimulus type (either moving checker board or static checker board). 

I don't think that we should add any further stimulation to the techniques, in fact actually delete the still existing one. We have a solution for that and it would be silly not to use it.

lzehl:
@UlrikeS91 we don't have a solution at the moment that fulfills all the needs of EBRAINS that is why we need to keep them here for now.

lzehl:
We can first look at a cleanup once a solution is mature. Fixing this will have to be postponed for now

lzehl:
let's assess this issue in an actual issue and prepare there for a clean solution that fits all needs (I'll try to set up the issue asap)

PR #303, opened 2022-09-14
Error fix - Pharmacology was in technique not in ExpApproach

UlrikeS91:
sloppy review by myself for #291...


lzehl (approved):
and sloppy PR from my side :see_no_evil: thanks for noticing and making the correction

lzehl:
@UlrikeS91 remind me that we need to do the cleanup for EBRAINS here

PR #308, opened 2022-09-19
Create setupType.schema.tpl.json

lzehl:


instances/terminology/setupType.jsonld
```diff
@@ -0,0 +1,15 @@
...
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/terminology/setupType",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Terminology",
+  "definition": "The setup type describes the overall purpose of arranging equipment in a certain way (setup).",
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "age category",
```
UlrikeS91: change to "setup type"

UlrikeS91 (changes requested):
I think the definition for computing system is fine. But there is a mistake that needs to be fixed. Otherwise, it's ok to merge from my side.

instances/terminology/setupType.jsonld
lzehl: @UlrikeS91 done! stupid mistake...

UlrikeS91 (approved):

Issue #312, opened 2022-09-22
New technique suggestion: movement tracking


spieschnik:
Hi,

Could you add "movement tracking" as a technique?

lzehl:
@UlrikeS91 we can make "motion tracking" as a synonym to "movement tracking"

UlrikeS91:
> @UlrikeS91 we can make "motion tracking" as a synonym to "movement tracking"

Motion tracking is already a synonym for the term (and the term was also already merged) 🙂

PR #313, opened 2022-09-23
Movement techniques

UlrikeS91:
related to #312 and additionally added a definition and synonyms for "eye movement tracking"

### DIFF
--- a/instances/technique/eyeMovementTracking.jsonld
+++ b/instances/technique/eyeMovementTracking.jsonld
@@ -4,7 +4,7 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/eyeMovementTracking",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": "'Eye movement tracking' is a technique to measure the eye movement and/or position to determine what an individual was focusing on at a given time.",
+  "definition": "'Eye movement tracking' refers to a group of techniques used to measure the eye movement and/or position of a living specimen over a given period of time.",
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
--- a/instances/technique/eyeMovementTracking.jsonld
+++ b/instances/technique/eyeMovementTracking.jsonld
@@ -6,10 +6,10 @@
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
   "definition": "'Eye movement tracking' refers to a group of techniques used to measure the eye movement and/or position of a living specimen over a given period of time.",
   "description": null,
-  "interlexIdentifier": null,
+  "interlexIdentifier": "http://uri.interlex.org/ilx_0417680",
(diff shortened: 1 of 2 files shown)


instances/technique/eyeMovementTracking.jsonld
```diff
@@ -4,11 +4,14 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/eyeMovementTracking",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": null,
+  "definition": "'Eye movement tracking' is a technique to measure the eye movement and/or position to determine what an individual was focusing on at a given time.",
```
lzehl: I would also refer this to a group of techniques. There is more than one to track eye movement and keep it general (without the purpose).

"'Eye movement tracking' refers to a group of techniques used to measure over a given period of time the eye movement and/or position of a living specimen."

instances/technique/eyeMovementTracking.jsonld
```diff
@@ -4,11 +4,14 @@
...
   "description": null,
   "interlexIdentifier": null,
```
lzehl: @tgbugs would this reference be correct? http://uri.interlex.org/base/ilx_0417680

instances/technique/movementTracking.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/movementTracking",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "'Movement tracking' includes a group of techniques to measure the movement and/or position of an individual or parts of an individual at a given time.",
```
lzehl: The term movement tracking is really generic. It does not refer only to subjects or body parts of a subject, but also to objects. I would probably use specimen here because it could also cover the movement of single cells / or small organisms

Suggestion:
"'Movement tracking' refers to a group of techniques used to measure over a given period of time the movement and/or position of an object, specimen, or anatomical parts of a specimen."

lzehl (changes requested):

instances/technique/eyeMovementTracking.jsonld
UlrikeS91: Ok, change it

instances/technique/eyeMovementTracking.jsonld
UlrikeS91: I added it, it sound correct to me. We can also fix it later in case it is wrong.

instances/technique/movementTracking.jsonld
UlrikeS91: True! Thanks for the updated definition.

lzehl (approved):

Issue #310, opened 2022-09-20
New experimental approach: biophysics


UlrikeS91:
@tgbugs noticed this is missing in a SPARC dataset context. We should also add it here.

PR #314, opened 2022-09-23
New exp. approach: biophysics

UlrikeS91:
#310 

@tgbugs any suggestion for a good definition?

### DIFF
--- a/instances/experimentalApproach/biophysics.jsonld
+++ b/instances/experimentalApproach/biophysics.jsonld
@@ -1,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/experimentalApproach/biophysics",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/ExperimentalApproach",
+  "definition": null,
+  "description": null,
+  "interlexIdentifier": null,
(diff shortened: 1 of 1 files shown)


instances/experimentalApproach/biophysics.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/experimentalApproach/biophysics",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/ExperimentalApproach",
+  "definition": null,
```
lzehl: @UlrikeS91 I would go with the general definition of wikipedia, and then later explain in description a bit more what specifically is considered to fall under biophysics.

Suggestion (@tgbugs please feel free to adapt and to suggest a more detailed description in addition):
"'Biophysics' covers any experimental approach that is traditionally used in physics, but applied and modified to study biological phenomena. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Biophysics)]"

lzehl (changes requested):

instances/experimentalApproach/biophysics.jsonld
UlrikeS91: Added your definition 🙂

lzehl (approved):

PR #326, opened 2022-10-14
Fix synonym case

ehennestad:


lzehl:
@ehennestad could you update this branch so that we only have the relevant changes in this PR? the previous PRs from you with the format corrections are accepted and merged.

lzehl (approved):

Issue #335, opened 2022-10-19
technique request: principal component analysis


lzehl:
[principal component analysis](https://en.wikipedia.org/wiki/Principal_component_analysis)

PR #336, opened 2022-10-19
Create principalComponentAnalysis.jsonld

lzehl:
@apdavison is the definition okay? related to #335


UlrikeS91 (approved):

UlrikeS91:
I think the definition is fine (and understandable), but I'm not an expert. Maybe we want to wait for @apdavison review before merging?

lzehl:
@UlrikeS91  I think for now that definition is okay. I'm anyway planning a hackathon where we can tackle definitions of analysis techniques with experts. For now they should just not be completely wrong

apdavison (approved):

PR #345, opened 2022-11-18
update action status types

lzehl:
@apdavison I'm adding the interlex IDs where it is available, added synonyms and adding 'inactive' as additional type.


apdavison (commented):
I don't think these identifiers are appropriate (see comment from @tgbugs on https://scicrunch.org/scicrunch/interlex/view/ilx_0100286).

I also don't think "inactive" is specific enough as a term, rather we could consider adding "paused", "on hold", "pending", although I would prefer just to stick to the https://schema.org/ActionStatusType enumeration for the sake of interoperability.

lzehl:
@apdavison the suggested interlex link was discussed with @tgbugs but I can remove it until the ontology issue is solved. 
I agree with your suggestion to split 'inactive' to 'paused', 'on hold' and 'pending'. But I also think we should keep 'inactive' as a broader term in addition.

I don't see why we should stick to only the schema.org action status types when they are clearly missing other action states (e.g. for inactive states). We still keep the interoperability with the one that they have.

Note that ActionStatusType may not only be needed for computational operations but can also be used in other contexts where action states need to be defined (e.g. actions in a behavioral task for example). I would not like to introduce an additional terminology for these types but rather have one (where maybe some terms are computer specific and others e.g. behavioral specific which should then be stated in the definition of the term).

apdavison:
"pending" would be useful. Otherwise, ok to merge.

instances/actionStatusType/completed.jsonld
```diff
@@ -4,11 +4,14 @@
...
-  "definition": "An action that has already taken place",
+  "definition": "An action that has already taken place.",
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
   "name": "completed",
   "preferredOntologyIdentifier": "https://schema.org/CompletedActionStatus",
-  "synonym": null
+  "synonym": [
+    "completed action status",
```
apdavison, line 14: I suggest adding "finished successfully"

instances/actionStatusType/completed.jsonld
```diff
@@ -4,11 +4,14 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/actionStatusType/completed",
   "@type": "https://openminds.ebrains.eu/controlledTerms/ActionStatusType",
-  "definition": "An action that has already taken place",
+  "definition": "An action that has already taken place.",
```
apdavison: I think some reference to the success of the action is needed, to distinguish it from "failed", which is also an action that has already taken place.

instances/actionStatusType/failed.jsonld
```diff
@@ -4,11 +4,14 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/actionStatusType/failed",
   "@type": "https://openminds.ebrains.eu/controlledTerms/ActionStatusType",
-  "definition": "An action that failed to complete",
+  "definition": "An action that failed to complete.",
```
apdavison: "or completed but produced an error"

instances/actionStatusType/failed.jsonld
```diff
@@ -4,11 +4,14 @@
...
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
   "name": "failed",
   "preferredOntologyIdentifier": "https://schema.org/FailedActionStatus",
-  "synonym": null
+  "synonym": [
```
apdavison, line 13: "error"

instances/actionStatusType/paused.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "@id": "https://openminds.ebrains.eu/instances/actionStatusType/paused",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/ActionStatusType",
+  "definition": "A temporarily stopped action that can be resumed at a later point in time.",
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "paused",
+  "preferredOntologyIdentifier": null,
+  "synonym": [
+    "paused action type",
```
apdavison, line 14: "suspended"

instances/actionStatusType/pending.jsonld
```diff
@@ -0,0 +1,18 @@
...
+  "@id": "https://openminds.ebrains.eu/instances/actionStatusType/pending",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/ActionStatusType",
+  "definition": "An action which is awaiting execution.",
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "pending",
+  "preferredOntologyIdentifier": null,
+  "synonym": [
+    "pending action type",
```
apdavison, line 16: "queued"

instances/actionStatusType/potential.jsonld
```diff
@@ -4,11 +4,14 @@
...
-  "definition": "A description of an action that is supported",
+  "definition": "A description of an action that is supported.",
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
   "name": "potential",
   "preferredOntologyIdentifier": "https://schema.org/PotentialActionStatus",
-  "synonym": null
+  "synonym": [
+    "completed action type",
```
apdavison: copy-paste error

apdavison (commented):
I have a few suggestions for additional synonyms and description changes.

instances/actionStatusType/completed.jsonld
lzehl: agreed. cf changes

instances/actionStatusType/failed.jsonld
lzehl: agreed. also "finished unsuccessfully"?

instances/actionStatusType/paused.jsonld
lzehl: agreed and added

instances/actionStatusType/pending.jsonld
lzehl: agreed and added

lzehl:
@apdavison I agreed with all of your suggestions and made the changes. Please double check again

PR #349, opened 2022-12-06
Show openMINDS logo in darkmode

ehennestad:


lzehl (approved):
I think this is missing now the anchor for the image, but if we do add it again we need the anchor changing with the mode...

PR #351, opened 2022-12-11
Adding three new techniques

eapapp:
Hello openMINDS team,

This is a proposal to add three new techniques (missing for the dataset I'm curating):

1. Electrocardiography
 - interlexIdentifier added
 - definition from Wikipedia (no def in interlex)

2. Galvanic skin response
 - interlexIdentifier not found
 - definition from [Nagai et al. 2019](https://www.frontiersin.org/articles/10.3389/fneur.2019.00377/full)

3. Photoplethysmography
 - interlexIdentifier added
 - definition from interlex

Please feel free to change them as you see fit. I hope they can be added soon, as we would like to publish the dataset before the holidays.

Thanks a lot,
Eszter


instances/technique/electrocardiography.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/electrocardiography",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Electrocardiography is the process of producing an electrogram of the heart: a graph of voltage versus time of the electrical activity of the heart using electrodes placed on the skin.",
+  "description": null,
+  "interlexIdentifier": "http://uri.interlex.org/base/ilx_0103701",
```
lzehl: this is not the same term. One is referring to the technique, the other to the protocol. Please set this one to null. We will integrate the term in the technique ontology and update your entry when an ID is available. @tgbugs FYI

instances/technique/electrocardiography.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/electrocardiography",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Electrocardiography is the process of producing an electrogram of the heart: a graph of voltage versus time of the electrical activity of the heart using electrodes placed on the skin.",
```
lzehl: suggestion to change to:

"An 'electrocardiography' is a non-invasive technique used to record the electrical activity of a heart using electrodes placed on the skin. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Electrocardiography)]"

Note: Please always add the reference from where you took / adapted the definition from if it is not your own phrasing.

instances/technique/electrocardiography.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "description": null,
+  "interlexIdentifier": "http://uri.interlex.org/base/ilx_0103701",
+  "knowledgeSpaceLink": null,
+  "name": "electrocardiography",
+  "preferredOntologyIdentifier": null,
+  "synonym": null
```
lzehl: I would add the following synonym: "ECG"

instances/technique/photoplethysmography.jsonld
```diff
@@ -0,0 +1,14 @@
...
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/photoplethysmography",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Plethysmographic determination in which the intensity of light reflected from the skin surface and the red cells below is measured to determine the blood volume of the respective area. There are two types, transmission and reflectance.",
+  "description": null,
+  "interlexIdentifier": "http://uri.interlex.org/base/ilx_0487650",
+  "knowledgeSpaceLink": null,
+  "name": "photoplethysmography",
+  "preferredOntologyIdentifier": null,
```
lzehl: Please add here "http://id.nlm.nih.gov/mesh/2018/M0026056" (according to InterLex)

instances/technique/photoplethysmography.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/photoplethysmography",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Plethysmographic determination in which the intensity of light reflected from the skin surface and the red cells below is measured to determine the blood volume of the respective area. There are two types, transmission and reflectance.",
```
lzehl: Suggestion to change this to:

"A 'photoplethysmography' is a non-invasive technique to optically detect blood volume changes in the micro-vascular bed of tissue by measuring the transmissive absorption and/or the reflection of light by/of the skin." 

@tgbugs I think the definition in InterLex should potentially be updated ?

instances/technique/photoplethysmography.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "description": null,
+  "interlexIdentifier": "http://uri.interlex.org/base/ilx_0487650",
+  "knowledgeSpaceLink": null,
+  "name": "photoplethysmography",
+  "preferredOntologyIdentifier": null,
+  "synonym": null
```
lzehl: Please add "PPG"

instances/technique/galvanicSkinResponse.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/galvanicSkinResponse",
```
lzehl: this is not a technique. The galvanic skin response which seems to be an outdated term for electrodermal activity (EDA) is a property that can be measured. EXAMPLE definition: EDA is an indirect measure of neurally mediated effects on sweat glands’ permeability, observed as changes in the resistance of the skin to a small electrical current or as differences in the electrical potential between different parts of the skin. ([source](https://www.analog.com/en/technical-articles/design-development-and-evaluation-of-a-system-to-obtain-electrodermal-activity.html))

For EBRAINS: please add this for now as term suggestion with the suggested terminology "BiologicalPhenomenon" (does not exist yet but is in prep). As for the technique: leave it out for now, but I will investigate if "epidermal electrophysiology technique" (or something like this) would be the correct superclass for any electrical measurement from the skin. @tgbugs & @UlrikeS91 TBD

lzehl (changes requested):
@eapapp thank you for those suggestions. Please see my requests for "electrocardiography" and "photoplethysmography". 

As for "galvanic skin response": this is not a technique, but a measurement. Please see my suggestions how you can solve this for the EBRAINS case right now.

instances/technique/electrocardiography.jsonld
eapapp: Done.

instances/technique/electrocardiography.jsonld
eapapp: I updated the definition. The word electrocardiography is typically used without an indefinite article in English when we talk about the technique in general and not a concrete case, unless coupled with another word, e.g. electrocardiography measurement, electrocardiography system, electrocardiography score etc. So I skipped the "an" at the beginning.

instances/technique/electrocardiography.jsonld
eapapp: Done.

instances/technique/electrocardiography.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/electrocardiography",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "Electrocardiography is a non-invasive technique used to record the electrical activity of a heart using electrodes placed on the skin. [adapted from Wikipedia]",
```
lzehl: You are still missing the actual reference as a hyperlink. Sorry if that was not clear from my example

instances/technique/electrocardiography.jsonld
lzehl: @eapapp that is totally fine. please set a hyperlink for the reference, though.

instances/technique/photoplethysmography.jsonld
eapapp: Done.

instances/technique/photoplethysmography.jsonld
eapapp: Done.

instances/technique/photoplethysmography.jsonld
eapapp: Def updated. Same note here, we don't need the indefinite article.

instances/technique/galvanicSkinResponse.jsonld
eapapp: galvanicSkinResponse.jsonld deleted from techniques. [Term suggestion](https://editor.kg.ebrains.eu/instances/eef5c3b4-03d9-4645-9d9a-0164f84ed40e) added.

lzehl (approved):

PR #357, opened 2022-12-20
split CranialWindowType

lzehl:
into...
CranialWindowConstructionType
CranialWindowReinforcementType


UlrikeS91 (changes requested):
- cranialWindowConstructionType: file names need to be changed according to name change
- terminology folder updates are missing (remove/replace ".../terminology/cranialWindowType.jsonld" with the 2 new ones

UlrikeS91 (approved):

Issue #354, opened 2022-12-14
new Technique: distortion correction


lzehl:
@skoehnen OR @ehennestad  I need help with the definition since distortion corrections are not only caused by lenses.

lzehl:
@apdavison please review the definition for this term. I hope it is adequate.

ehennestad:
Image distortion correction is the process of removing or reducing distortion in digital images caused by factors such as lens aberrations, camera motion, or non-ideal image acquisition conditions. This can be achieved through various techniques such as geometric correction, which adjusts for lens distortion, or motion correction, which compensates for camera shake. The goal of image distortion correction is to improve the visual quality and accuracy of the image. (chatGPT)

lzehl:
@ehennestad thanks. This definition is a bit incomplete since it does not take distortion due to the image perspective into account. The PR is linked. I think the full hierarchy (not all terms are yet in) would be:

+ distortion correction
    + image distortion correction (I would stick to this one for now; add the other ones later)
        + optical image distortion correction (OR lens distortion correction)
        + perspective image distortion correction
        + motion image distortion correction (?)
   + audio distortion correction (?)
   + etc. (?)

lzehl:
this issue is solved with #370

PR #370, opened 2023-01-15
Create imageDistortionCorrection.jsonld

lzehl:
related to #354


skoehnen (approved):
Looks good to me and the definition is general to cover all the cases I could think of.

Issue #346, opened 2022-11-28
remove GitHub and GitLab instance from FileRepositoryType


lzehl:
https://github.com/HumanBrainProject/openMINDS_controlledTerms/tree/v1/instances/fileRepositoryType

Remove "GitHub" and "GitLab" instance and add instead "git" (which is the type of the version control system that both web services are using). Add also "git-annex" as instance.

skoehnen:
I will take a look.

skoehnen:
Someone please check if it is correct.

lzehl:
@skoehnen could you also add git-annex to the list? (I assume that this makes sense) and double check the definitions of the other terms for file repo types? (so that they really give a hint on the peculiarities of the repo type?)

lzehl:
I will close this one now and create a new issue for missing FileRepoTypes

PR #374, opened 2023-01-15
Consolidation of different git repository type into one unified instance

skoehnen:
As requested in #346


lzehl:
@skoehnen thank you for taking this over. What about git-annex though? and I think we have to review the definitions of the other terms in this terminology (so that they describe the peculiarities of the repository type)

instances/fileRepositoryType/git.jsonld
```diff
@@ -0,0 +1,21 @@
...
+    "preferredOntologyIdentifier": null,
+    "synonym": [
+        "Git repository",
+        "Git remote repository",
+        "remote Git repository",
+        "GitHub repository",
+        "GitLab repository"
+    ]
+  }
+
```
lzehl, line 21: newline missing at the end of the jsonld

skoehnen:
> @skoehnen thank you for taking this over. What about git-annex though? and I think we have to review the definitions of the other terms in this terminology (so that they describe the peculiarities of the repository type)

I would keep git-annex separate, while it share the same mechanics to store version-information, the handling of data is quite different.

lzehl:
@skoehnen what I mean is could you please also create an instance for git-annex (since it is also a repository type) ?

skoehnen:
> @skoehnen what I mean is could you please also create an instance for git-annex (since it is also a repository type) ?

@lzehl I will start a new PR for git-annex, can we merge this pull-request?

apdavison (approved):

lzehl (approved):

Issue #376, opened 2023-01-17
add biochemistry and culturing to exp approach


lzehl:
discussed and agreed upon with @tgbugs @UlrikeS91 @ehennestad

PR #377, opened 2023-01-17
new ExpApproaches: biochemistry, culturing

lzehl:
related to #376 

biochemistry: @tgbugs not sure what to add as Interlex ID (seems you have two?)
culturing: @tgbugs does not yet exist in Interlex correct?

Both definitions might need polishing.


instances/experimentalApproach/biochemistry.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/experimentalApproach/biochemistry",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/ExperimentalApproach",
+  "definition": "'Biochemistry' covers any experimental approach that studies chemical processes within and relating to living organisms. [adapted from [wikipedia](https://en.wikipedia.org/wiki/Biochemistry)]",
```
tgbugs: For consistency I would call it `biochemical approach` with `biochemistry` as a synonym to avoid conflation of the field of study with the experimental approach.

instances/experimentalApproach/culturing.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/experimentalApproach/culturing",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/ExperimentalApproach",
+  "definition": "'Culturing' covers any experimental approach that focuses on breeding, care, and keeping of living things in order to harvest the substances they produce.",
```
tgbugs: That definition is closer to husbandry. I don't have a good definition off the top of my head here, will think on it.

tgbugs (changes requested):
Probably also want to align the defintions with the usual conventions that we come up with in the pan projects sheet.

instances/experimentalApproach/biochemistry.jsonld
lzehl: @tgbugs this would be in conflict with how we named all experimental approaches in openMINDS (e.g., histology, electrophysiology etc). In principle we use a mixed interpretation in openMINDS (something between field of study and exp. approach).

But I definitely should add biochemical approach to the synonyms (same for culturing)

instances/experimentalApproach/culturing.jsonld
lzehl: thanks. I really wasn't sure what you all want to include here.

lzehl:
@tgbugs any update?

instances/experimentalApproach/culturing.jsonld
```diff
@@ -0,0 +1,16 @@
...
```
tgbugs:
```suggestion
  "definition": "A culturing approach includes any experimental approach that involves sustaining single cells or extracted parts of multicellular organisms outside of their natural environment.",
```

lzehl (commented):
@aeidi89 I've updated the definition and changed the term name / id / file name for those approaches already to solve the issue raised by @tgbugs about mixing "field of study" with "experimental approach".

We can fix all other experimental approaches in that regard in a follow up PR. So feel free to merge this PR if it is okay.

aeidi89 (approved):
Looks good to me.

Issue #352, opened 2022-12-12
adding a super class for epidermal electrophysiology techniques


lzehl:
@tgbugs & @UlrikeS91 I need your input here.

There are a bunch of different epidermal electrophysiology te)chniques of which some received a clear term (e.g. electrocardiography) and some not (e.g. measurement of electrodermal activity). For the latter it is typically just said that electrodes were placed on the skin of a subject...

Could we discuss a proper technique term for the latter case?

This issue came up in the PR #351 by @eapapp

UlrikeS91:
I thought that we actually discussed this issue before for some reason. I think it makes sense to add a superclass. I think "epidermal electrophysiology technique" would be suitable.

lzehl:
top. I will add it asap

lzehl:
question remains if we find a proper term for the case "electrodes were placed on the skin of a subject"

lzehl:
vs. the proper techniques EKG, EEG, etc which are also "epidermal electrophysiology technique"

UlrikeS91:
We don't have any techniques (yet?) that represent the act of placing electrodes (or similar devices). I'm not sure we would need them, though. 
The placement of the electrode on the skin is a requirement to be able to do epidermal ephys. Some goes for other placements, like inserting an electrode into a cell before doing intracellular injections or the insertion of a biopsy needle into the liver to do a liver biopsy. 
I know that this is a bit oversimplified, but I don't think we need to cover the placement unless that is something extremely special or is the only act (as in "placing the electrode but then not doing anything with it").

PR #380, opened 2023-01-20
Create epidermalElectrophysiologyTechnique.jsonld

lzehl:
related to #352 

@UlrikeS91 the definition does not make me happy yet... I'm open for suggestions ;)

### DIFF
--- a/instances/technique/epidermalElectrophysiologyTechnique.jsonld
+++ b/instances/technique/epidermalElectrophysiologyTechnique.jsonld
@@ -4,7 +4,7 @@
   },
   "@id": "https://openminds.ebrains.eu/instances/technique/epidermalElectrophysiologyTechnique",
   "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
-  "definition": "The term 'epidermal electrophysiology technique' refers to any electrophysiology technique where electrodes are placed onto the skin of an organism to non-invasively (or minimally invasively) measure electrical functions of it's body.",
+  "definition": "The term 'epidermal electrophysiology technique' describes a subclass of non-invasive electrophysiology techniques where one or several electrodes are placed on the outermost cell layer of an organism (epidermis) to measure electrical properties.",
   "description": null,
   "interlexIdentifier": null,
   "knowledgeSpaceLink": null,
--- a/instances/technique/epidermalElectrophysiologyTechnique.jsonld
+++ b/instances/technique/epidermalElectrophysiologyTechnique.jsonld
@@ -1,0 +1,18 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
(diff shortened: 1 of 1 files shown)


instances/technique/epidermalElectrophysiologyTechnique.jsonld
```diff
@@ -0,0 +1,18 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/technique/epidermalElectrophysiologyTechnique",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/Technique",
+  "definition": "The term 'epidermal electrophysiology technique' refers to any electrophysiology technique where electrodes are placed onto the skin of an organism to non-invasively (or minimally invasively) measure electrical functions of it's body.",
```
UlrikeS91: Suggestion: 
"'Epidermal electrophysiology technique' describes a subclass of electrophysiology techniques where one or several electrodes are placed on the outermost cell layer of an organism (epidermis) to measure electrical properties." 

(probably still room for improvements...)

instances/technique/epidermalElectrophysiologyTechnique.jsonld
```diff
@@ -0,0 +1,18 @@
...
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "epidermal electrophysiology technique",
+  "preferredOntologyIdentifier": null,
+  "synonym": [
+    "cutaneous electrophysiology",
+    "cutaneous electrophysiology technique",
```
UlrikeS91: Is this correct? These seem more like related terms and not exact synonyms. "cutaneous" is only "related to the skin", "epidermal" is more specific.

UlrikeS91 (changes requested):

instances/technique/epidermalElectrophysiologyTechnique.jsonld
lzehl: not sure. This article seems to use it as a synonym: https://www.nature.com/articles/s41551-019-0365-8 What do you think?

instances/technique/epidermalElectrophysiologyTechnique.jsonld
lzehl: updated with small changes. thank you!

instances/technique/epidermalElectrophysiologyTechnique.jsonld
UlrikeS91: Yeah, I also came across that one but they only use cutaneous once in the first sentence. I got the impression that they were speaking of cutaneous techniques in general. My interpretation was that epidermal techniques are a subclass of cutaneous techniques, but I could be wrong 😉

lzehl: don't know either. let's delete them (we can always add them again)

UlrikeS91 (approved):

Issue #83, opened 2021-11-16
adding molecularEntity


lzehl:
@tgbugs & @UlrikeS91 I'm surprised that no one requested that yet... we postponed this for a while.
do you agree adding the schema and at least some instances?

[@UlrikeS91 did you prepare something for this already? I have the feeling you did... if so, feel free to forward, I can take over if you want]

UlrikeS91:
Yes, I have some first instances for a molecularEntity terminology ☺️ I have a list that includes more than just those, so it's not smart to link it here. But we can discuss the list offline and I can make a PR with what we landed on/want to include (which can be up for feedback from the broader community afterwards instead). 

Sounds good?

tgbugs:
I'm guessing the domain is CHEBI + protein ontology/uniprot, so small molecules and proteins? Would this also cover engineered proteins such as eGFP, RFP, etc. which might have certain metadata templates about excitation frequencies or similar?

UlrikeS91:
@tgbugs, here is an excerpt from the list. "Keywords from EBRAINS KG" are terms that have been stated as keywords in KG v2, then I did a rough "categorization" stating a potential terminology the term could fit in (here you only got the terms that could have something to do with molecular entities), "openMINDS (potential) name" states the (re)curate name of the term (work in progress, which is why several of them are still empty), "InterLex" states the InterLex IRI (if I could find one) and "preferred ontology ID" states either the IRI that has been stated as preferred in InterLex or, if I couldn't find it in InterLex, another ontology. 
When a term wasn't included in InterLex, my first "go-to" was CHEBI. In some cases, I also used the CHEBI as preferred ID even though InterLex stated something else as preferred (e.g. for chemical agents like isoflurane or DAB). A few also point to the protein ontology (e.g. Neuroligin 3) but I think those are mostly cases where the protein ontology was listed in InterLex.

eGFP, RFP and such haven't been used as keywords in the KG. Closest to those are probably terms like "Alexa Fluor 594" or "FluorRuby", which I would include in the molecularEntity terminology for openMINDS. So, I would probably add terms like eGF/RFP/etc. as well. 
The original idea for this terminology was to not restrict it too much and collect anything that loosely related to it, and rather later on split it up into several sub-terminologies (with stricter definitions). 


keyword from EBRAINS KG | categorization | openMINDS (potential) name | InterLex | preferred ontology ID
-- | -- | -- | -- | --
[...]

PR #383, opened 2023-01-30
Create isometricMapping.jsonld

lzehl:


lzehl:
@mdenker could you double check the definition of this technique? thanks in advance

mdenker:
Though I'm no expert here, in my understanding, this entry is still unprecise. I would define isometric mapping, in general, as the process of mapping a source space onto a target space, such that the distance between any source points is equal to that between mapped target points. In the special case of low-dimensional embedding of a measurement space using a non-linear technique, isometric mapping would refer to techniques that likewise preserve -- to some degree -- the distances in the low-dimensional space. In this respect I find that ISOMAP (the wikipedia article) would represent one particular type of such isometric mappings. For example, Multi-dimensional Scaling would be another such technique that approximately preserves distances. Thus, depending on what should be modelled here, I would either rename the method to ISOMAP if that specific technique is meant, or broaden the definition to be less specific about the algorithm.

lzehl:
@mdenker thank you a lot for the feedback. Indeed the ISOMAP is for now the issue. I will also add isometric mapping as the super class. I'll pin you again to double check the definitions.

UlrikeS91:
Replaced by PR https://github.com/openMetadataInitiative/openMINDS_instances/pull/83

PR #389, opened 2023-02-08
Create trachemysScriptaElegans.jsonld

apdavison:
currently curating a model of the cerebellum of this species


lzehl (approved):
I've made some small adjustments

Issue #394, opened 2023-02-17
Add new disease term: Multiple sclerosis


UlrikeS91:
Requested by one of our team members here in Oslo. Just need to track it, so we don't forget about it.

aeidi89:
Thank you for the suggestion. Yes, we can integrate this.

lzehl:
solved with #395

PR #395, opened 2023-02-21
Create multipleSclerosis.jsonld

aeidi89:
Add new disease term: Multiple sclerosis #394


lzehl (approved):
@aeidi89 thanks for making this PR and taking care of the issue

PR #397, opened 2023-02-21
Create testIR.jsonld

ingrre:


lzehl (commented):
@ingrre thanks for the test. Seems to work :)

Issue #390, opened 2023-02-09
New 'celltype' suggestion


szafarnia:
Can "granular neuron" be registered as a new celltype?

lzehl:
@szafarnia yes this can be added as a super class term.

name: "granule neuron"
definition: "The term 'granule neuron' refers to a set of neuron types typically found in granular layers across brain regions whose only common feature is that they all have very small cell bodies. [adapted from [Wikipedia](https://en.wikipedia.org/wiki/Granule_cell)]"
synonym: "granule cell"

@tgbugs this would be a super class I think to a couple of cell types already registered in InterLex. Would you agree with name, definition and synonym?

tgbugs:
Hrm, yeah, we can probably add such a neuron. I think I had it defined at one point but we removed it. I can add it to npo.

lzehl:
TODO, add CellType instance:
name: "granule neuron"
definition: "The term 'granule neuron' refers to a set of neuron types typically found in granular layers across brain regions whose only common feature is that they all have very small cell bodies. [adapted from [Wikipedia](https://en.wikipedia.org/wiki/Granule_cell)]"
synonym: "granule cell"

(links to interlex can be added once the term is available there as well)
@aeidi89 or @spieschnik would you be interested taking this over?

spieschnik:
Yes, I can do the PR

apdavison:
Note that we already have "granule cell" - https://github.com/HumanBrainProject/openMINDS_controlledTerms/blob/v1/instances/cellType/granuleCell.jsonld

lzehl:
@apdavison thank you... I didn't realize. @spieschnik then we do not have to do this PR. Double check though the entry and add the information that is missing. We need to double check as well if we follow the policy we decided on for registering cell types. I think the entry might need to be changed to "granule neuron" with synonym "granule cell"...

lzehl:
@szafarnia instance is updated. You should find it by now in EBRAINS

PR #401, opened 2023-03-01
Update and rename granuleCell.jsonld to granuleNeuron.jsonld

spieschnik:
#390 
made changes as suggested in the issue


aeidi89:
@lzehl I added you as a reviewer and marked this PR with the 'CLEAN-UP for v3 release'. Is the label correct?

lzehl (approved):
@spieschnik thanks for this PR I've made a small correction (you forgot to rename the name ;))

lzehl:
@spieschnik OR @aeidi89 TODO: this PR causes a cleanup action in EBRAINS. All references to the old term "granuleCell" need to be replaced with the reference to the new term "granuleNeuron". Once every references to old term "granuleCell" are removed, the old term "granuleCell" needs to be deleted from the KG.

PR #404, opened 2023-03-02
Create isHomologousTo.jsonld

lzehl:
"is identical to" is too restrictive in describing, e.g., cross-species relations; I suggest to use "is homologous to" for such cases.


aeidi89 (approved):
Looks good

Issue #420, opened 2023-05-23
new technique ? HiP-CT


lzehl:
Hirachical i? phase contrast tomography

lzehl:
https://mecheng.ucl.ac.uk/hip-ct/

lzehl:
solved with #426

PR #426, opened 2023-06-29
Create hierarchicalPhaseContrastTomography.jsonld

aeidi89:
#420


lzehl (approved):
@aeidi89 I only deleted one additional space in the definition, deleted the abbreviation from the definition (we have them listed structurally as synonyms), and got rid of the capitalization of the term (in definition and name).

Issue #378, opened 2023-01-19
new disease terms: tumor, cancer


lzehl:
We need the superclasses for all tumor types and cancer.

UlrikeS91:
I think that we should add:

**"neoplasia" (http://purl.obolibrary.org/obo/MONDO_0023370):**
- with "neoplastic disease" and "neoplastic disorder" as synonyms
- definition: "A benign or malignant tissue growth resulting from uncontrolled cell proliferation." or "'Neoplasia' is the uncontrolled, abnormal growth of cells or tissue in the body resulting in a neoplasm or tumor (benign or malignant)."
- Note: "neoplasm" (http://purl.obolibrary.org/obo/MONDO_0005070, which could have "tumor" or "mass" as synonyms) is the result, it's not typically classified as the disease, but we could add "neoplasm" instead if this is easier to understand

**"cancer" (http://purl.obolibrary.org/obo/MONDO_0004992):**
- is a subclass of neoplasia/neoplasm
- definition: "A malignant tissue growth resulting from uncontrolled cell proliferation." or "'Neoplasia' is the uncontrolled, abnormal growth of cells or tissue in the body resulting in a malignant neoplasm or tumor."

This one was a nice source: http://publichealth.lacounty.gov/vet/docs/CancerAnEng.pdf

lzehl:
I have the feeling that MONDO is not very consistent in the definitions / hierarchy of the ontology in this regard. Should maybe use this source with care.

I would also classify tumor more as the result (specification) of the disease. So your suggestions make sense to me. 
But we also registered already glioma and meningioma in diseases as well... 
So for me neoplastic disorder includes benign or malignant tumors. And then each of those have different types. 

Let's discuss this in the next ontology meeting.

UlrikeS91:
> I have the feeling that MONDO is not very consistent in the definitions / hierarchy of the ontology in this regard. Should maybe use this source with care.
> 
> I would also classify tumor more as the result (specification) of the disease. So your suggestions make sense to me. But we also registered already glioma and meningioma in diseases as well... So for me neoplastic disorder includes benign or malignant tumors. And then each of those have different types.
> 
> Let's discuss this in the next ontology meeting.

[...]

PR #427, opened 2023-06-29
Create cellProliferationDisorder.jsonld

aeidi89:
#378


lzehl (approved):
@aeidi89 according to discussion in #378 this term has no synonyms (so far). I deleted them.

PR #437, opened 2023-07-17
Create imageSequence.jsonld

apdavison:
similar to "naturalMovie" but less natural


lzehl (approved):
I think I know what you mean. We should add definitions to those terms as soon as possible

Issue #438, opened 2023-08-10
New Disease


archgogo:
name: trigeminal neuralgia
definition: Trigeminal neuralgia is sudden, severe facial pain.
interlex identifier: https://neuinfo.org/interlex/view/ilx_0758102?searchTerm=Trigeminal%20neuralgia

lzehl:
@archgogo thanks for requesting this. It is in process #439

aeidi89:
@archgogo #439 has now been approved and I'll close this issue.

PR #439, opened 2023-08-14
Create trigeminalNeuralgia.jsonld

lzehl:
related to #438


aeidi89 (approved):
This looks like a perfect example of a controlled term.

Issue #442, opened 2023-08-18
new analysis technique: Hilbert transform?


lzehl:
@apdavison is this an analysis technique or just the result of one?

apdavison:
An analysis technique

lzehl:
cf. #447

UlrikeS91:
will be fixed in https://github.com/openMetadataInitiative/openMINDS_instances/pull/85

PR #85, opened 2021-11-17
Create molecularEntity Terminology

lzehl:
related to #83 

Include steps:
1) create controlledTerms schema
2) add molecularEntity to terminology list
3) add first set of instances for molecularEntity

@UlrikeS91 let me know if I missed something.

PR #447, opened 2023-08-30
Create HilbertTransformation.jsonld

lzehl:
related to #442


apdavison (changes requested):
Looks fine, but should be "transform", not "transformation". The term "Hilbert transformation" is not used in English.

UlrikeS91:
replaced by https://github.com/openMetadataInitiative/openMINDS_instances/pull/85

PR #458, opened 2023-11-17
Molecular Entities needed for in depth 

Peyman-N:


instances/molecularEntity/2-Bromo-2-chloro-1,1,1-trifluoroethane.jsonld
```diff
@@ -0,0 +1,14 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/molecularEntity/2-Bromo-2-chloro-1,1,1-trifluoroethane",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/MolecularEntity",
+  "definition": "2-Bromo-2-chloro-1,1,1-trifluoroethane is a nonflammable, halogenated, hydrocarbon anesthetic that provides relatively rapid induction with little or no excitement. ",
```
lzehl: please delete the unnecessary space at the end of the definition and add the source of the definition (e.g. as citation and/or link to the source)

instances/molecularEntity/2-Bromo-2-chloro-1,1,1-trifluoroethane.jsonld
```diff
@@ -0,0 +1,14 @@
...
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "2-Bromo-2-chloro-1,1,1-trifluoroethane",
+  "preferredOntologyIdentifier": "https://pubchem.ncbi.nlm.nih.gov/compound/87224111",
+  "synonym": []
```
lzehl: if there are no synonyms this should be null.

instances/molecularEntity/DNQX.jsonld
```diff
@@ -0,0 +1,17 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/molecularEntity/DNQX",
```
lzehl: please follow one convention for chemical entities: I would suggest we always use the IUPAC  name.

instances/molecularEntity/DNQX.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "@type": "https://openminds.ebrains.eu/controlledTerms/MolecularEntity",
+  "definition": "DNQX (6,7-dinitroquinoxaline-2,3-dione) is a competitive antagonist at AMPA and kainate receptors, two ionotropic glutamate receptor (iGluR) subfamilies.",
```
lzehl: Please add the source of the definition. Adapt according to the decision made for the naming convention. Do never list synonyms in the definition. We should always just list them under "synonym".

instances/molecularEntity/DNQX.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "description": "It is used in a variety of molecular biology subfields, notably neurophysiology, to assist researchers in determining the properties of various types of ion channels and their potential applications in medicine.",
```
lzehl: Please add the source of the description. Please adapt so that the description can be read without the definition.

instances/molecularEntity/DNQX.jsonld
lzehl: I would also delete the further information on the receptors which is irrelevant for this entity. I would though add it's antagonistic competitors for the receptors (and or clearly identify which binding site it uses).

instances/molecularEntity/DNQX.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "interlexIdentifier": "http://uri.interlex.org/base/ilx_0103368",
+  "knowledgeSpaceLink": null,
+  "name": "DNQX",
```
lzehl: see naming conventions suggestion: IUPAC Name

instances/molecularEntity/DNQX.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "preferredOntologyIdentifier": "https://pubchem.ncbi.nlm.nih.gov/compound/3899541",
+  "synonym": [
```
lzehl: correct to change in naming convention

instances/molecularEntity/EDTA.jsonld
```diff
@@ -0,0 +1,30 @@
...
+    "@context": {
+        "@vocab": "https://openminds.ebrains.eu/vocab/"
+    },
+    "@id": "https://openminds.ebrains.eu/instances/molecularEntity/EDTA",
+    "@type": "https://openminds.ebrains.eu/controlledTerms/MolecularEntity",
+    "definition": "Ethylenediaminetetraacetic acid (EDTA), also called edetic acid after its own abbreviation, is an aminopolycarboxylic acid with the formula [CH2N(CH2CO2H)2]2.",
+    "description": "This white, water-soluble solid is widely used to bind to iron (Fe2+/Fe3+) and calcium ions (Ca2+), forming water-soluble complexes even at neutral pH. It is thus used to dissolve Fe- and Ca-containing scale as well as to deliver iron ions under conditions where its oxides are insoluble. EDTA is available as several salts, notably disodium EDTA, sodium calcium edetate, and tetrasodium EDTA, but these all function similarly.",
+    "interlexIdentifier": null,
+    "knowledgeSpaceLink": "https://knowledge-space.org/wiki/CHEBI:4735#ethylene-glycol-bis-2-aminoethyl-tetraacetic-acid",
+    "name": "EGTA",
```
lzehl: wrong name. but should be anyway changed to IUPAC Name

instances/molecularEntity/EDTA.jsonld
```diff
@@ -0,0 +1,30 @@
+{
+    "@context": {
+        "@vocab": "https://openminds.ebrains.eu/vocab/"
+    },
+    "@id": "https://openminds.ebrains.eu/instances/molecularEntity/EDTA",
+    "@type": "https://openminds.ebrains.eu/controlledTerms/MolecularEntity",
+    "definition": "Ethylenediaminetetraacetic acid (EDTA), also called edetic acid after its own abbreviation, is an aminopolycarboxylic acid with the formula [CH2N(CH2CO2H)2]2.",
```
lzehl: source?

lzehl (changes requested):
@Peyman-N thank you for this PR. I know that adding these entities is urgent but I would like to establish some conventions for adding chemical entities.

I would suggest the following:
+ name: use the IUPAC nomenclature
+ atID suffix: use the IUPAC name
+ filename: use the IUPAC name
+ definition: always cite your source; do not list synonyms here; I would always include here though the molecular formula
+ description: always cite your source; do not list synonyms here
+ preferredOntologyID: do not use pubchem (for what I can see this is not an ontology); use ChEBI instead
+ synonyms: only use exact synonyms; list alphabetically

Please let me know if you would discuss this suggestions or, when you agree with them, need help updating this PR.

apdavison:
Do we really want IUPAC names in all cases? For example, "sucrose" is a much more useful name than "β-D-fructofuranosyl α-D-glucopyranoside"

Since we anyway list synonyms, and provide ontology links where possible, I think we should use the name that most scientists would recognize as the "name" property _and_ in the `@id`, so "DNQX" not "6,7-dinitroquinoxaline-2,3-dione"

On further investigation, using IUPAC names in `@id`s will often be problematic because the names contain characters that cause problems in IRIs (need to be encoded, systems _may_ but are not _required_ to accept them as valid IRIs), such as square brackets, e.g.

"magnesium;[[[(2R,3S,4R,5R)-5-(6-aminopurin-9-yl)-3,4-dihydroxyoxolan-2-yl]methoxy-hydroxyphosphoryl]oxy-oxidophosphoryl] hydrogen phosphate" (MgATP)

lzehl:
@apdavison we need to discuss (I see the issue with IUPAC as well). However using trivial names causes issues as well because not every component has one and we end up being very inconsistent within the terminology. I would like to have something consistent as name / id and then put everything else in the synonyms

apdavison:
This is a nice comparison of the major public chemistry databases: https://doi.org/10.1002/cmdc.201700724

As a naming convention, I suggest picking one of them, let's say PubChem, since it has a better search interface than UniChem, ChemSpider is apparently not fully public, and ChEBI is missing a lot of terms.

For the first few chemicals in this PR, this would give the following (PubChem name, followed by current name in this PR in parentheses):

- halothane (2-Bromo-2-chloro-1,1,1-trifluoroethane)
- 6,7-Dinitro-1,4-dihydroquinoxaline-2,3-dione (DNQX)
- Edetic Acid (EDTA)
- Egtazic acid (EGTA)
- 4-(2-Hydroxyethyl)-1-piperazine ethanesulfonic acid (HEPES)
- Magnesium ATP (MgATP)

lzehl:
@apdavison thank you for doing some research to solve this. I think your suggestion is great. Let's go with PubChem. We should check if we could get in contact with them long-term (to see if we can collaborate/contribute directly)

lzehl:
@Peyman-N all updates done or are you still on it? I'm meeting with @tgbugs now and will mention our decision. Hope he agrees.

lzehl:
@Peyman-N @apdavison feedback from @tgbugs : they are using ChEBI in the ontology (NIF/InterLex) but PubChem is good as well. If terms are missing from ChEBI @tgbugs would be interested to know which terms these are. @apdavison do you have examples?

apdavison:
> do you have examples?

none to hand. This comment is based on the review article I mentioned above, and on ChEBI having less comprehensive lists of synonyms (for example, "DNQX" gives no hits in ChEBI, but if you try search for it in PubChem then use the synonyms you find there to search ChEBI you can find it).

Peyman-N:
Hello, everyone.
Sorry for the late response.

When it comes to small molecules, ChEBI is comprehensive enough for our purposes. The only small molecule that I couldn't find is [Silver ammonium](https://pubchem.ncbi.nlm.nih.gov/compound/57440423). However, for larger molecules, it becomes less extensive, especially for drugs and proteins.

Anyway, I really prefer **ChEBI** as it is more widely used by the community. On the other hand, PubChem anchors to multiple ontologies, which is nice. However, I think we should discuss the possibility of anchoring to multiple ontologies in the future.

lzehl:
@Peyman-N @apdavison  what does this mean now? you would like to connect use the ChEBI name when available and to PubChem when it is not available in ChEBI? OR do we stick to original suggestion to only use the PubChem name? (@Peyman-N connecting to multiple ontologies is handled by InterLex, I would not like to take on this task as well)

Peyman-N:
No no, sorry I didn't explain myself correctly. I mean for naming convention we would use PubChem.
I was speaking about prefer ontologie, any way the interlex and PubChem, would take care of the linkage there.

lzehl:
got it @Peyman-N thanks for clarifying!

apdavison (approved):

apdavison:
@lzehl can we merge this?

lzehl:
@apdavison yes please. one approval should be enough. We do not need to have two people approving for instances (that were discussed beforehand) and smaller issues I would say.

apdavison:
@lzehl I agree in general, but since this PR also has the goal of establishing conventions I think we should all approve it.

lzehl:
@apdavison as we approved / agreed on a convention in the discussion I merged it now already because I thought this one is urgent. I can go over the instances maybe some time this week in detail and do corrections if needed.

PR #461, opened 2023-11-24
Adding new units of measurement 

Peyman-N:


instances/unitOfMeasurement/hertz.jsonld
```diff
@@ -0,0 +1,18 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/unitOfMeasurement/hertz",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement",
+  "definition": "Hertz is the unit of frequency in the International System of Units (SI), equivalent to one event (or cycle) per second.",
```
lzehl: The hertz is ... AND please add the source of that definition (here is is Wikipedia).

instances/unitOfMeasurement/kilohertz.jsonld
```diff
@@ -0,0 +1,17 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/unitOfMeasurement/kilohertz",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement",
+  "definition": "A unit of frequency equal to one thousand cycles per second; one thousand hertz.",
```
lzehl: please match form of definition used for hertz

instances/unitOfMeasurement/kilovolt.jsonld
```diff
@@ -0,0 +1,17 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/unitOfMeasurement/kilovolt",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement",
+  "definition": "One thousand volts.",
```
lzehl: please match the definition form you used for hertz/kilohertz (here naturally vold / kilovolt)

instances/unitOfMeasurement/microvolt.jsonld
```diff
@@ -0,0 +1,17 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/unitOfMeasurement/microvolt",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement",
+  "definition": "One millionth (10^-6) of a volt.",
```
lzehl: please adapt (see previous comments)

instances/unitOfMeasurement/nanovolt.jsonld
```diff
@@ -0,0 +1,16 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/unitOfMeasurement/nanovolt",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement",
+  "definition": " one billionth (10^-9) of a volt.",
```
lzehl: correct (unnecessary space) and see above

lzehl (changes requested):
We should discuss conventions here. I'm unsure if my suggestions are the best solution for definitions.

Peyman-N:
I completely agree we should standardize the definition all of the SI prefixes. 

I would suggest this definition :

A unit of `measurand` equal to `amount of` `main unit`.
Example:
A unit of **electric potential** equal to **_one thousand_** **volts**.

Or we can go the hard way and separate the SI prefixes from the unit, this can be beneficial in long run but would need a clean up; so I am against it.

apdavison:
> A unit of `measurand` equivalent to `amount of` `main unit`.

"equal to", not "equivalent to"

instances/unitOfMeasurement/kilovolt.jsonld
```diff
@@ -0,0 +1,17 @@
...
+  "@type": "https://openminds.ebrains.eu/controlledTerms/UnitOfMeasurement",
+  "definition": "A unit of electric potential equal to one thousand volts.",
+  "description": null,
+  "interlexIdentifier": null,
+  "knowledgeSpaceLink": null,
+  "name": "kilovolt",
+  "preferredOntologyIdentifier": "http://purl.obolibrary.org/obo/UO_0000248",
+  "synonym": [
+    "kV",
+    "KV"
```
apdavison: I don't think we should list such non-standard forms as "KV"

apdavison (changes requested):
one small change: remove "KV" from synonyms

lzehl (approved):
under the condition to remove KV from synonyms of kilovolt as @apdavison suggests.

apdavison (approved):

PR #466, opened 2023-12-19
Create hertz

apdavison:


Peyman-N:
The hertz is already present in pull request #461, adding it two times would cause a conflict.

apdavison:
sorry, I forgot about #461.

PR #482, opened 2024-10-21
Pittsburgh Compound B (PiB)

eapapp:
Hello,

I suggest to add Pittsburgh Compound B (PiB), a PET tracer that binds to beta-amyloid plaques, to controlledTerms/molecularEntity. Haven't found it on InterLex or in the KnowledgeSpace, is that a problem? Not sure what to add under preferredOntologyIdentifier either. What do you think?

Best,
Eszter

### DIFF
--- a/instances/molecularEntity/pittsburghCompoundB.jsonld
+++ b/instances/molecularEntity/pittsburghCompoundB.jsonld
@@ -1,0 +1,20 @@
+{
+  "@context": {
+    "@vocab": "https://openminds.ebrains.eu/vocab/"
+  },
+  "@id": "https://openminds.ebrains.eu/instances/molecularEntity/pittsburghCompoundB",
+  "@type": "https://openminds.ebrains.eu/controlledTerms/MolecularEntity",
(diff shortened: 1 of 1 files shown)


lzehl:
@eapapp thanks for making this suggestion. However could you move this PR to the [openMINDS_instances](https://github.com/openMetadataInitiative/openMINDS_instances) repository (into folder v3.0)? We do not integrate instances anymore into the submodules. The instances prior collected here all moved to the new openMINDS_instances repo and the instances here in the submodule will soon be deleted.

For you suggestion: molecularEntity is definitely the right terminology library. Please use this ontology as reference: https://pubchem.ncbi.nlm.nih.gov/compound/2826731
Please use also the name: (N-methyl-(11c))2-(4'-methylaminophenyl)-6-hydroxybenzothiazole
and add Pittsburgh Compound B, Pittsburgh B, PiB as synonyms

eapapp:
Hi @lzehl, thanks for your help. I updated the file and moved it over to https://github.com/openMetadataInitiative/openMINDS_instances/pull/176.

PR #487, opened 2025-02-06
MRI acquisition type

Peyman-N:
Adding MRI acquisition type to controlled terms.
Derived from [DICOM tag (0018,0023)](https://dicom.nema.org/medical/Dicom/2024c/output/chtml/part03/sect_C.8.3.html):
Instances:

- 1D

- 2D
      frequency x phase

- 3D
      frequency x phase x phase


schemas/MRAcquisitionMode.schema.tpl.json
lzehl, line 1: Please also update the filename

lzehl (changes requested):
@Peyman-N I don't have a strong opinion here. Is MR acquisition type more used than MRI acquisition mode ?

Peyman-N:
@lzehl Sorry, I thought I had changed all of them to acquisition type. Originally, after our meeting, I wanted to name it mode, but then I reviewed the literature and they are almost always acquisition type.

lzehl (approved):

lzehl:
@Peyman-N please add also the terminology instance for this new controlledTerms schemas

PR #502, opened 2025-08-14
Introduce Jurisdiction related terminologies.

Raphael-Gazzotti:
Controlled terms for https://github.com/openMetadataInitiative/openMINDS_core/issues/513

### DIFF
--- a/schemas/country.schema.tpl.json
+++ b/schemas/country.schema.tpl.json
@@ -1,4 +1,0 @@
-{
-  "_type": "controlledTerms:Country",
-  "_extends": "controlledTerm.schema.tpl.json"
-}
--- a/schemas/country.schema.tpl.json
+++ b/schemas/country.schema.tpl.json
@@ -1,0 +1,4 @@
+{
+  "_type": "controlledTerms:Country",
+  "_extends": "controlledTerm.schema.tpl.json"
+}
--- a/schemas/geopoliticalGroup.schema.tpl.json
+++ b/schemas/geopoliticalGroup.schema.tpl.json
@@ -1,7 +1,0 @@
-{
-  "_type": "controlledTerms:GeopoliticalGroup",
-  "_extends": "controlledTerm.schema.tpl.json",
-  "_categories": [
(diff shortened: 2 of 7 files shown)


lzehl:
We may want to have the following for covering jurisdiction/geopolitical coverage under the category "jurisdiction":
- SovereignState  (synonym: country)
- SupranationalBody (e.g. European Union)
- SubnationalEntity (dependent jurisdictions within a country)
- SubnationalAutonomousEntity (synonym: territory; independent jurisdictions within a country)
- GeopoliticalGroup (synonym: region; geographical groups with no common political system; e.g. Europe)

SubnationalEntity and SubnationalAutonomousEntity we could also leave out for now because there are a lot and they are less important for our current use case

apdavison:
Why do we need such fine-grained detail? This is going to be painful to use in practice. My feeling is that "address" as a string is sufficient for scientific purposes. If people need to do more fine-grained analysis, they can parse the strings.

lzehl:
@apdavison address as a string will not work (not reliable for getting the information we need and would even then require code to interpret). we need explicitly the jurisdiction for several cases (e.g. legal home of organisations, ethics jurisdiction for datasets). 

@Raphael-Gazzotti I would say we do not need SubnationalEntity for the current cases. We could potentially leave out the GeopoliticalGroup for now (because strictly speaking these are not having any legal jurisdiction power).

lzehl:
@Raphael-Gazzotti reduce to SovereignState (synonym: country), SupranationalBody (e.g. European Union)

lzehl (changes requested):
Please remove: 
- LegalEntityType (different PR after discussing on https://github.com/openMetadataInitiative/openMINDS_controlledTerms/issues/507)
- SubnationalAutonomousEntity
- SubnationalEntity
- GeopoliticalGroup

lzehl (approved):

lzehl:
@Raphael-Gazzotti @apdavison we need to remember to add those schemas as terminologies to Terminology

PR #505, opened 2025-09-20
Add InterfaceType

apdavison:
cf https://github.com/openMetadataInitiative/openMINDS_core/issues/527

### DIFF
--- a/schemas/interfaceType.schema.tpl.json
+++ b/schemas/interfaceType.schema.tpl.json
@@ -1,0 +1,4 @@
+{
+  "_type": "controlledTerms:InterfaceType",
+  "_extends": "controlledTerm.schema.tpl.json"
+}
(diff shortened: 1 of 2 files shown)


schemas/interfaceType.schema.tpl.json
```diff
@@ -0,0 +1,4 @@
+{
+  "_type": "controlledTerms:InterfaceType",
```
lzehl: We should discuss the schema name. Maybe this is too ambiguous? Not sure maybe InteractionInterfaceTypes ? This could then also include VUI besides, GUI, API and CLI

lzehl: change type name to: CommunicationInterfaceType

lzehl: @apdavison changed the name but please confirm if this was the one we agreed to

apdavison: yes, I confirm

lzehl (approved):

PR #510, opened 2025-11-04
Introduce OrganizationType.

### LINKED ISSUE #507: new controlledTerm: LegalEntityType
extracted from https://github.com/openMetadataInitiative/openMINDS_core/issues/513
 
Linked type for new property "type" on organizations.

Question: Should we rename this to Entity Legal Forms (ELF) which is the official name of that terminology defined by  Global Legal Entity Identifier Foundation (GLEIF) under ISO 20275?

This is an extensive list because it is jurisdiction specific but we won't have to maintain it, since it is provided by GLEIF (see [ELF list](https://www.gleif.org/en/lei-data/code-lists/iso-20275-entity-legal-forms-code-list); even with official Latin-script rendering).

@Raphael-Gazzotti this should go into its own PR later.

Raphael-Gazzotti:
Closes #507

### DIFF
--- a/schemas/legalEntityType.schema.tpl.json
+++ b/schemas/legalEntityType.schema.tpl.json
@@ -1,0 +1,4 @@
+{
+  "_type": "controlledTerms:LegalEntityType",
+  "_extends": "controlledTerm.schema.tpl.json"
+}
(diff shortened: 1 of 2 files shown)


lzehl (approved):
@apdavison would you agree to the name change from LegalEntityType to OrganizationType?

Issue #260, opened 2022-08-18
new technique: Baseline measures


szafarnia:
Dear all,

can „**Baseline measures**” be added as a technique to the KG? Please find below corresponding description that I received from a data provider.

Best,
Sara

**Description:**
From “Määttä, J. I., van den Bosch, R., Papadopetraki, D., Hofmans, L., Lambregts, B., Westbrook, A., Verkes, R.-J., & Cools, R. Predicting effects of methylphenidate and sulpiride on brain and cognition: A pharmaco-fMRI, PET study. Design and descriptives. Open Science Framework. https://doi.org/10.31219/osf.io/d3h8e (2021)”:
“Various questionnaires and assessments were administered off-drug across the study to establish participants (a) working memory capacity (Listening span & digit span; Daneman & Carpenter, 1980, Groth-Marnat, 2001, Listening span and digit span, respectively), (b) crystalized intelligence (NLV/NART, Crawford, Stewart, Cochrane, Parker, & Besson, 1989; Schmand et al., 1991), (c) fluid intelligence (WAIS-IV-NL, Wechsler, 1997), (d) creativity (Kaufman, 201s2), (e) Need for Cognition (Cacioppo et al., 1984), behavioural inhibition and activation (Franken et al., 2005) (Carver & White, 1994), (f) trait impulsivity (Patton et al., 1995) and spontaneous eye blink rate (Groman et al., 2014) …. Except for the WAIS-IV-NL matrix test and the Dutch reading test (NLV/NART) all tasks were computerized. Due to an experimenter error a proto version of the trait impulsivity assessment (i.e., BIS-11A) was administered (administered in 2017-2018). After approval from the ethics committee the intended version (i.e., BIS-11) was sent out to the participants after the completion of the study, 67 out of the 100 participants completed the assessment…. 
Spontaneous eye blink rate (sEBR) was always acquired before 5 PM over a period of up to 10-minutes, using two vertical and two horizontal Ag-AgCl electrodes placed around the eyes. Participants were not aware of their eye blink rate being recorded. During the course of the study sEBR acquisition time were increased from 6 to 10 minutes (28 versus 72 participants, respectively), motivated by recommendation of external colleagues…. Furthermore, 10 participants received inconsistent and inaccurate instructions for the spontaneous eye blink rate acquisition (they were instructed to stare at the wall)….”

UlrikeS91:
[...]

PR #511, opened 2025-11-05
Include accessibility-controlled terminologies

Raphael-Gazzotti:
Add accessibility-controlled terminologies from [issue #260](https://github.com/openMetadataInitiative/openMINDS_instances/issues/260)
, including AccessMode, AccessType, AuthorizationProcess, and PaymentType.

Note: A cleanup of ProductAccessibility may be needed once all terminologies, schemas are in place.

### DIFF
--- a/schemas/accessProcessType.schema.tpl.json
+++ b/schemas/accessProcessType.schema.tpl.json
@@ -1,4 +1,4 @@
 {
-  "_type": "controlledTerms:AuthorizationProcess",
+  "_type": "controlledTerms:AccessProcessType",
   "_extends": "controlledTerm.schema.tpl.json"
 }
--- a/schemas/paymentModel.schema.tpl.json
+++ b/schemas/paymentModel.schema.tpl.json
@@ -1,4 +1,4 @@
 {
-  "_type": "controlledTerms:PaymentType",
+  "_type": "controlledTerms:PaymentModel",
   "_extends": "controlledTerm.schema.tpl.json"
 }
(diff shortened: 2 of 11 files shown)


schemas/accessProcessType.schema.tpl.json
lzehl, line 1: I think this should be AuthorizationProcessType (because we just conceptualize the type of process here and not describing the acutal process)

lzehl (changes requested):
@apdavison @olinux it would be good to get your feedback here as well.

schemas/accessProcessType.schema.tpl.json
Raphael-Gazzotti: Should this be renamed to AccessProcessType? I'm basing this on our last discussion on [Issue #260](https://github.com/openMetadataInitiative/openMINDS_instances/issues/260#issuecomment-3522869946).

schemas/paymentModelType.schema.tpl.json
lzehl, line 1: Maybe should better be named PaymentModel. From what I see payment type is rather associated with how you pay (e.g. credit card) and payment model with how you get charged (e.g. one-time, recurring, usage-based, etc)

lzehl (changes requested):

schemas/paymentModel.schema.tpl.json
```diff
@@ -0,0 +1,4 @@
+{
+  "_type": "controlledTerms:PaymentModel",
```
lzehl:
```suggestion
  "_type": "controlledTerms:PaymentModelType",
```

schemas/paymentModelType.schema.tpl.json
lzehl, line 1: paymentModelType.schema.tpl.json

lzehl (changes requested):

lzehl (commented):
updated according to recent discussions in https://github.com/openMetadataInitiative/openMINDS_instances/issues/260 on Accessibility schema

lzehl (approved):

PR #512, opened 2025-11-12
Introduce modificationClause.schema.tpl.json

Raphael-Gazzotti:
Needed for https://github.com/openMetadataInitiative/openMINDS_core/issues/505


lzehl:
@Raphael-Gazzotti after going over this again I think this should be called ModificationClause (used in contractual context) not ModificationTerms (used in IP context).

lzehl (approved):

PR #516, opened 2026-01-15
Add legalCountry property to organizationType.schema.tpl.json

### LINKED ISSUE #515: Add sovereign state context to OrganizationType
The current OrganizationType schema does not capture the sovereign state in which an OrganizationType is legally defined.

Extending the schema to include this information would allow organization types to be constrained by the legal context of an organization, reducing ambiguity and preventing confusion between similarly named organization types that exist under different sovereign states.

Raphael-Gazzotti:
Closes #515

### DIFF
--- a/schemas/organizationType.schema.tpl.json
+++ b/schemas/organizationType.schema.tpl.json
@@ -1,13 +1,12 @@
 {
-  "required": [
-    "legalCountry"
-  ],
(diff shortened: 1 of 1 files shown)


lzehl:
after discussing again with @Raphael-Gazzotti we are closing this and minimizing the OrganizationType to legalEntity and organizationalUnit for now

